PR #8, opened 2024-05-02
Use capitalised clever references

pitmonticone:
- [x] Substitute `Theorem \ref` with `\Cref`
- [x] Substitute `Lemma \ref` with `\Cref`
- [x] Substitute `Prop \ref` with `\Cref`
- [x] Substitute `Proposition \ref` with `\Cref`
- [x] Substitute `Corollary \ref` with `\Cref`

`\ref`s included in \chapter, \section, etc. are excluded in order to avoid `hyperref` warning such as: 

> Package hyperref Warning: Token not allowed in a PDF string (Unicode):
(hyperref)                removing `\new@ifnextchar' on input line 1019.

### DIFF
--- a/blueprint/src/chapter/main.tex
+++ b/blueprint/src/chapter/main.tex
@@ -30,7 +30,7 @@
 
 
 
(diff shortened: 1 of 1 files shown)


fpvandoorn:
This is useful, but is not good while the "real" source code is on Overleaf. Please do this again once this becomes the official source code in 1-2 weeks.

pitmonticone:
Oh, right. Sorry for this...

PR #22, opened 2024-06-08
Update blueprint section 10

ldiedering:
Various rewrites in the blueprint to change from piecewise constant to smooth approximation, as already done in the Lean version. 
Many other small changes and corrections to sync blueprint and Lean code, in particular some more \leanok's have been added. 
However, I realized that the paper has changed regarding the "local oscillation" (this term does not even appear any more) and that the Lean code needs some updates of definitions to reflect this change before I can add more \leanok's to things from subsection 10.9 .


fpvandoorn:
Thanks! I'll work on the existing Lean code in the upcoming week

PR #36, opened 2024-06-23
Added metric space Carleson variant for two-sided Calderon-Zygmund kernel

ldiedering:
The previous chapter 10 is split into two parts. 
The first one is a variant of metric space Carleson for a two-sided Calderon-Zygmund kernel and where the assumption on T* is replaced with a weaker one (for an operator without the two suprema). 
The second one then uses this new variant to prove classical Carleson. 

This changes mainly required to rewrite the previous 10.3 and 10.5 to the general case of doubling metric measure spaces.


blueprint/src/print.tex
```diff
@@ -29,7 +29,10 @@
 
 \usepackage{amsthm}
 \usepackage{etexcmds}
-\usepackage{thmtools}
+% Commented out thmtools because in combination with cleveref it currently leads to broken links,
+% e.g. 'Lemma 3' will always point to Lemma 3 of Chapter 1 and not necessarily the actually referred one. See also
+% https://tex.stackexchange.com/questions/598167/cleveref-sorting-doesnt-work-correctly-if-theorem-shares-subsection-counter
+% \usepackage{thmtools}
```
fpvandoorn, line 35: This should be fixed in the latest online version. If you agree, please revert.

fpvandoorn (commented):
I'll take a look over the tex changes later.

fpvandoorn:
Great! Thanks!

PR #48, opened 2024-06-30
Enforce `a`-related parameters as natural numbers

Parcly-Taxel:
More specifically `a, A = 2 ^ A, D = 2 ^ (100 * a ^ 2)` and the constant `S` in the definition of `GridStructure`.

The blueprint explicitly states that `a` is natural, and having this avoids "rounding fudges" in some proofs. Cf. [Zulip](https://leanprover.zulipchat.com/#narrow/stream/442935-Carleson/topic/Rounding.20errors.20in.20the.20proofs/near/448087018).

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -96,21 +96,23 @@ jobs:
           sudo chown -R runner docs
           cp -r .lake/build/doc docs/docs
 
-      - name: Bundle dependencies
-        uses: ruby/setup-ruby@v1
-        with:
-          working-directory: docs
-          ruby-version: "3.0" # Not needed with a .ruby-version file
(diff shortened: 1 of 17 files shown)


fpvandoorn (commented):
Looks good to me, thanks! two comments:
* Can you revert the changes to `DoublingMeasure`? A doubling measure is a concept that I would like to move to Mathlib relatively soon, and there is no reason to restrict the parameter to a natural number abstractly.
* You might want to cast the natural numbers to real numbers as soon as possible, by explicitly writing e.g. `(A : ℝ) ^ 2`. That might save some `exact_mod_cast`s.

fpvandoorn:
Thanks!

PR #49, opened 2024-06-30
Add first version of proofs for task 15

jim-portegies:
Task 15: Fill in proofs in first part of `WeakType.lean`


fpvandoorn:
Thanks! `continuousWithinAt_distribution` looks quite difficult indeed. I hoped there were some related Mathlib results that would make it much easier...

PR #50, opened 2024-07-02
Lemma 2.1.2 and wiggle order lemmas

Parcly-Taxel:
This also removes all but one of the `exact_mod_cast`s left over from the naturalisation of `a, A, D, S`.


fpvandoorn:
Thanks!

PR #56, opened 2024-07-04
Lemmas 5.2.3, 5.2.4, 5.3.4

Parcly-Taxel:
Also extract a common proof motif in `GridStructure`.


fpvandoorn:
Thanks! This looks great!

PR #59, opened 2024-07-07
Lemma 5.2.2

Parcly-Taxel:
I had to add the property `coeGrid_measurable` of `GridStructure` in order to use `measure_biUnion_finset`, but this property is specified in the blueprint:

> A grid structure $(\mathcal{D}, c, s)$ on $X$ consists of a finite collection $\mathcal{D}$ of pairs $(I, k)$ of **Borel sets** in $X$...


fpvandoorn:
Thanks! And thanks for adding the missing field to `GridStructure`!

PR #60, opened 2024-07-07
Proof of Lemma 2.1.3 and the other results in Psi.lean

js2357:
I had to add assumptions (in particular, `IsCZKernel a K`) in a few theorems. There were also some places where the theorems included unnecessary assumptions (`hs : s ∈ Icc (-S : ℤ) S`), which I removed.

The blueprint says that `X` is a metric space, but the lean file made it a `PseudoMetricSpace` instead. I left it as a `PseudoMetricSpace` because this did not make the proofs substantially more difficult, but if desired `X` could be changed to a metric space to simplify the proofs a little bit.

The code is kind of slow. I tried to speed it up a bit, but someone with more experience might be able to do better.


fpvandoorn:
Thanks, this is great! We can try to speed up the code later. I forgot to add `IsCZKernel a K` to `ProofData`. I will do that myself.

PR #67, opened 2024-07-09
Proposition 2.0.1

js2357:
Prove `finitary_carleson` (Proposition 2.0.1).

Also clean up a few apparent typos:

* There were several instances of `Measurable G'` which I changed to `MeasurableSet G'`.
* I changed the definition of `C2_0_1` so that it's defined to be equal to `C2_0_2`. (The constants in the blueprint are not equal, but this appears to be a typo.)
* I changed the definition of `C2_0_2` from `C5_1_2 a q + C5_1_2 a q` to `C5_1_2 a q + C5_1_3 a q`.
* I changed `1/q` in the statement of `finitary_carleson` to `q⁻¹` for consistency with `discrete_carleson`.


fpvandoorn:
Oh great! That was easier than I expected :-)
All typo fixes seem great to me (and I'm very much in favor of defining the constants in terms of previous constants!)

PR #68, opened 2024-07-11
Prove integrable_tile_sum_operator

js2357:
Prove `integrable_tile_sum_operator`.

As an intermediate step, prove that `y ↦ Ks s x y` is integrable.


fpvandoorn:
Thanks!

PR #73, opened 2024-07-12
Prove Icc_subset

js2357:
Prove the lemma that I was previously calling `Icc_subset`. (I hadn't noticed that the `S` in FinitaryCarleson.lean was notation for `S X`.)

Rename the lemma `Icc_σ_subset_Icc_S` and move it to Defs.lean, alongside related lemmas.


fpvandoorn:
Yeah, I was a bit worried whether all my local notation obscured the meaning of some variables.

Thanks!

PR #74, opened 2024-07-13
Lemma 5.4.(1–3)

Parcly-Taxel:


fpvandoorn:
If you want, I can merge it like this, or a version where we `sorry` the symmetry case for now.

Parcly-Taxel:
> If you want, I can merge it like this, or a version where we `sorry` the symmetry case for now.

Nah – I've finished the proof of 5.4.2!

PR #79, opened 2024-07-16
Lemmas 5.4.5 to 5.4.7

Parcly-Taxel:


fpvandoorn:
Great! Also good that you're separating out some lemmas.

PR #80, opened 2024-07-16
Speed up CI 

pitmonticone:
This PR ensures correct docs caching and minimization of disk space usage
- [x] Fix "Cache mathlib documentation" step (from 12 to 2 minutes)
- [x] Fix "Remove lake files from documentation" (remove from GitHub page but not from cache: 378MB to 369MB)

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -57,6 +57,7 @@ jobs:
             .lake/build/doc/Std
             .lake/build/doc/Mathlib
             .lake/build/doc/declarations
+            .lake/build/doc/find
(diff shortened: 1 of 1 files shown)


fpvandoorn:
Thanks!

We should keep an eye out that the "Make sure the cache works"-step is indeed not necessary (I think it is indeed unnecessary, since we're copying from `.lake/docs` instead of moving the file).

pitmonticone:
> I think it is indeed unnecessary, since we're copying from `.lake/docs` instead of moving the file.

Yes, exactly.

PR #81, opened 2024-07-17
Lemma 5.2.7

Parcly-Taxel:


fpvandoorn:
I'll merge this as-is, but feel free to clean it up in a follow-up PR, e.g. by
* Make `X` explicit in `layervol` (`variable (X) in`)
* Feel free to keep extra lemmas about `layervol`/`stackSize` around
* Since `lintegral_Ioc_partition` is in the `ToMathlib` folder, it should probably be generalized a bit before it is Mathlib-ready. It would be nice to have it stated for the case where you generalize 
`∑ l ∈ Finset.Ico a b, ∫⁻ t in Ioc (l * c) ((l + 1 : ℕ) * c), f t` to something like `∑ l ∈ Finset.Ico a b, ∫⁻ t in Ioc (g l) (g (Order.succ l)), f t` where and `g` is a monotone function.

PR #87, opened 2024-07-19
Added some \lean's and \leanok's to the blueprint, split proofs in 10.2 and 11.6.

ldiedering:


fpvandoorn:
Thanks!

fpvandoorn:
Were the new comments at the bottom of `CarlesonOperatorReal` intentional?

PR #88, opened 2024-07-20
Prove Lemma 5.2.8

js2357:
Prove Lemma 5.2.8.

Notes:

* I changed the statement of Lemma 5.2.8 to use `ℝ` instead of `ℕ`, and changed the exponent `9 * a - j` to use subtraction in  `ℤ` instead of `ℕ`. I think the lemma is only used to prove an inequality in `ℝ≥0`, so this shouldn't cause any problems.
* I corrected a minor error in the blueprint's proof, where it referred to `u` and `u'` as possible elements of `B'`, rather than `𝒬 u` and `𝒬 u'` (and then needed an extra injectivity argument to finish the proof).


fpvandoorn:
Thanks!

PR #89, opened 2024-07-21
Prove lemmas about wnorm

js2357:
Prove the lemmas about `wnorm` and `MemWℒp` in WeakType.lean.

`HasStrongType.hasWeakType` was originally stated with the assumption `(hp : 1 ≤ p)`, which needed to be changed to `(hp' : 1 ≤ p')`.

I also corrected very minor typos in the documentation of `HasWeakType` and `HasStrongType`.


fpvandoorn:
Thanks!

PR #97, opened 2024-07-29
Definition of `𝓩` and its properties

Parcly-Taxel:
Completes Task 9.


fpvandoorn:
Thanks!

Btw, please leave the empty line between the proof of a lemma and the end of a section (so that it is clear that the `end` is not part of the proof).

PR #106, opened 2024-08-03
Lemma 5.5.1 and 5.5.3

Parcly-Taxel:
I had to make one changes to the blueprint, both of which appear to be typos:
* The equality of `antichain_decomposition` is weakened to the forwards inclusion, because I could not prove the reverse inclusion. This appears to be enough, looking at the proof of `forest_complement`.


Parcly-Taxel:
@fpvandoorn

fpvandoorn:
I'm quite sure that the previous definition of `𝔓₁` is correct. I'll ask my coauthors how to deal with `𝔏₄` in the proof of 5.5.1.

fpvandoorn:
Actually, I looked a bit more myself. If `p ∈ 𝔓pos` you can conclude that `¬ (𝓘 p : Set X) ⊆ G'` hence `¬ (𝓘 p : Set X) ⊆ G₃`, hence `p ∉ 𝔏₄ k n j`

PR #107, opened 2024-08-04
Prove MeasureTheory.SublinearOn.maximalFunction

js2357:
- The definition of `SublinearOn` was missing the requirement `c ≥ 0`.
- The statement of `MeasureTheory.SublinearOn.maximalFunction` had a spurious `p`; `MB` is defined with p=1, so there's no need for a general `p`.
- The blueprint says that `𝓑` is finite, so I used that assumption even though several theorems in the file assume `𝓑` is countable instead. I also changed the assumption in `hasStrongType_MB`.

### DIFF
--- a/Carleson/ToMathlib/Misc.lean
+++ b/Carleson/ToMathlib/Misc.lean
@@ -159,6 +159,27 @@ lemma lintegral_Ioc_partition {a b : ℕ} {c : ℝ} {f : ℝ → ℝ≥0∞} (hc
       Nat.Ico_succ_right_eq_insert_Ico h, Finset.sum_insert Finset.right_not_mem_Ico,
       add_comm (lintegral ..), ih]
 
+/-! ## Averaging -/
+
+-- Named for consistency with `lintegral_add_left'`
+-- Maybe add laverage/setLaverage theorems for all the other lintegral_add statements?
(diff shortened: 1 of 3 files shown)


Carleson/ToMathlib/Misc.lean
```diff
@@ -185,6 +185,27 @@ lemma lintegral_Ioc_partition {a b : ℕ} {c : ℝ} {f : ℝ → ℝ≥0∞} (hc
       Nat.Ico_succ_right_eq_insert_Ico h, Finset.sum_insert Finset.right_not_mem_Ico,
       add_comm (lintegral ..), ih]
 
+/-! ## Averaging -/
+
+-- Named for consistency with `lintegral_add_left'`
+-- Maybe add laverage/setLaverage theorems for all the other lintegral_add statements?
+lemma setLaverage_add_left' {α : Type*} {m0 : MeasurableSpace α} {μ : MeasureTheory.Measure α}
```
fpvandoorn, line 192:
```suggestion
lemma setLAverage_add_left' {α : Type*} {m0 : MeasurableSpace α} {μ : MeasureTheory.Measure α}
```
also elsewhere

fpvandoorn:
Thanks!
I was hoping we could generalize this to countable `𝓑`, but it's no problem if we cannot.

PR #115, opened 2024-09-06
Proof of the real interpolation theorem

jim-portegies:
Proof of the real interpolation theorem. The proof roughly follows the exposition in Folland's book, but there are some changes as well, mainly to get a(n) (nicer) expression for the constant, and because in some respects it gives a simpler argument. Note also that the proof works for `NormedAddCommGroups`, and doesn't require the spaces to have a linear structure, and that only subadditivity, rather than sublinearity, is assumed for `T`, and finally that the proof works for constants `p0 p1 q1 q0 > 0`.

There is also a way in which the theorem is weaker than the one that was originally presented in the file, and this is on the aspect of preserving measurbility of `T`: `T` was previously assumed to preserve measurability on the union of `Lp0` and `Lp1`, and now it is assumed to preserver measurability on `Lp`. I do believe that this is unavoidable.

Please note that the order of the arguments in the final constant has changed. Of course, if this is a problem, I can change them back.


Carleson/RealInterpolation.lean
```diff
@@ -1,31 +1,1893 @@
...
-  [NormedAddCommGroup E₃] --[NormedSpace ℝ E₃] [FiniteDimensional ℝ E₃]
+variable {p₀ q₀ p₁ q₁ p q : ℝ≥0∞} {t : ℝ}
+
+/-! ## Convenience results for working with (interpolated) exponents
+-/
+namespace ComputationsInterpolatedExponents
+
+lemma preservation_positivity {p q : ℝ} (hp : p > 0) (hq : q > 0) (ht : t ∈ Ioo 0 1) :
+  0 < (1 - t) * p⁻¹ + t * q⁻¹ :=
+  add_pos (mul_pos (sub_pos.mpr ht.2) (inv_pos_of_pos hp))
```
b-mehta, line 40: This is a super minor comment, but it looks like the `positivity` tactic will help with some of these goals

pitmonticone: Yes, this will work:

```suggestion
  add_pos (mul_pos (sub_pos.mpr ht.2) (by positivity)) (mul_pos ht.1 (by positivity))
```

I have tested it locally, so you can safely commit it directly.

b-mehta: This isn't quite what I mean - instead I mean something like `cases ht; positivity` for the entire proof, instead of just a part of it.

fpvandoorn: I believe `t < 1 -> `1 - t > 0` is intentionally outside the scope of `positivity`, so I think many of these proofs subgoals are too complicated for `positivity`.

fpvandoorn:
Thanks, this is great! The extra hypothesis is no problem at all, and the generalization is nice!

I invite anyone to help clean-up these proofs.

PR #128, opened 2024-09-15
Miscellaneous golfs, mostly to RealInterpolation.lean

grunweg:
- compress some apply/exact statements when a clear win
- use fun_prop and positivity more
- (twice) extract an auxiliary lemma from an argument
- remove some `have`s which are used at most once or twice
- use .le instead of le_of_lt
- remove one adaptation note which is now obsolete
- use dot notation more


grunweg:
Sorry for the size: I can surely break this into several pieces (but only in a week).

pitmonticone:
Thank you very much @grunweg! 

I'll take a look at it later today.

pitmonticone:
Looks good to me. Thanks again!

PR #133, opened 2024-09-25
Update mathlib, deduplicate some lemmas

Parcly-Taxel:


fpvandoorn:
I merged master, so that CI now checks whether doc-gen is up-to-date. 
If CI now fails, please update mathlib using the command `lake update -Kenv=dev` or using `scripts/update-mathlib.sh`.

PR #136, opened 2024-09-26
Finish the proof of Lemma 7.1.5 + Lemma 7.4.7

Parcly-Taxel:
…and along the way tweak the definition of `upperRadius` to be `ENNReal`-valued, to accommodate the possibility of the relevant set being unbounded.


fpvandoorn:
Thanks!

PR #145, opened 2024-10-16
LeanAgent Proofs

Adarsh321123:
[LeanAgent](https://arxiv.org/abs/2410.06209) discovers proofs for theorems with the _sorry_ keyword.


fpvandoorn:
Thanks!
This was already proven on `master`, but the LeanAgent Proof was slightly shorter/nicer, so I used your proof.

PR #153, opened 2024-10-26
Task 100

erdOne:
Fill in the sorry for Lemma 7.6.4 (square function count).
Modify the statement to match the blueprint. 
Give a better upper bound and updated the blueprint accordingly.

### DIFF
--- a/blueprint/src/chapter/main.tex
+++ b/blueprint/src/chapter/main.tex
@@ -5840,7 +5840,7 @@ $$
         For each $J \in \mathcal{J}'$ and all $s \ge 0$, we have
         $$
             \frac{1}{\mu(J)} \int_J \Bigg(\sum_{\substack{I \in \mathcal{D}, s(I) = s(J) - s\\ I \cap \scI(\fu_1) = \emptyset\\
-        J \cap B(I) \ne \emptyset}} \mathbf{1}_{B(I)}\bigg)^2 \, \mathrm{d}\mu \le 2^{104a^2} (8 D^{-s})^\kappa\,.
+        J \cap B(I) \ne \emptyset}} \mathbf{1}_{B(I)}\bigg)^2 \, \mathrm{d}\mu \le 2^{14a+1} (8 D^{-s})^\kappa\,.
         $$
     \end{lemma}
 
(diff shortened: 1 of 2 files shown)


blueprint/src/chapter/main.tex
```diff
@@ -5852,18 +5852,18 @@ \section{Proof of The Remaining Tiles Lemma}
...
     $$
-        \mu(B(x, 9D^{s})) \ge \sum_{I \in \mathcal{D}_{s,x}} \mu(B(c(I), \frac{1}{4} D^{s(I)})) \ge 2^{-8a} |\mathcal{D}_{s,x}| \mu(B(x, 9D^{s}))\,.
+        \mu(B(x, 9D^{s})) \ge \sum_{I \in \mathcal{D}_{s,x}} \mu(B(c(I), \frac{1}{4} D^{s(I)})) \ge 2^{-7a} |\mathcal{D}_{s,x}| \mu(B(x, 9D^{s}))\,.
     $$
     Dividing by the positive $\mu(B(x, 9D^{s}))$, we obtain that for each $x$
     \begin{equation}
         \label{eq-sep-small-bound}
         \Bigg(\sum_{\substack{I \in \mathcal{D}, s(I) = s(J) - s\\ I \cap \scI(\fu_1) = \emptyset\\
-        J \cap B(I) \ne \emptyset}} \mathbf{1}_{B(I)}(x) \bigg)^2 = |\mathcal{D}_{s(J) - s,x}|^2 \le 2^{16a} \,.
+        J \cap B(I) \ne \emptyset}} \mathbf{1}_{B(I)}(x) \bigg)^2 \le |\mathcal{D}_{s(J) - s,x}|^2 \le 2^{14a} \,.
```
erdOne, line 5866: I don't think we can prove equality because there might be some element in $\mathcal{D}_{s(J) - s,x}$ such that $I \cap \mathscr{I}(u_1)$ is non-empty?

We only need the upperbound though.

erdOne:
Should I go through the lemmas downstream of this in the blueprint and update the bound accordingly? 
I don't have the confidence to do this mistake-free though.

fpvandoorn:
This is great! Thanks

> Should I go through the lemmas downstream of this in the blueprint and update the bound accordingly? I don't have the confidence to do this mistake-free though.

No, that is not necessary. We can do that as we formalize those lemmas.

PR #163, opened 2024-11-08
Add `create-release` workflow

pitmonticone:
Adds a `create-release` workflow which automates the creation of a new Git tag and GitHub release when the `lean-toolchain` file is updated in the `main` branch.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,0 +1,66 @@
+name: Create Release
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2  # Fetch the last two commits for comparison
+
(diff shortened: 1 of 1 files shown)


fpvandoorn:
The default branch is called `master` in this repository. Does that mean the script requires a fix?

Other than that, feel free to treat this as a `bors d+`, i.e. feel free to merge yourself.

fpvandoorn:
Actually, I agree with Ruben here: https://leanprover.zulipchat.com/#narrow/channel/416277-FLT/topic/versioning
I think this is only really needed/desired for stable Mathlib versions, although I don't care that much either way.

pitmonticone:
Yes, it requires a substitution! Thanks.

pitmonticone:
Done.

pitmonticone:
> Actually, I agree with Ruben here: https://leanprover.zulipchat.com/#narrow/channel/416277-FLT/topic/versioning I think this is only really needed/desired for stable Mathlib versions, although I don't care that much either way.

I agree with Ruben too. That's why I have added `stable` to the `rev` argument of mathlib in #164.

PR #167, opened 2024-11-11
chore: enable two uncontroversial mathlib linters

grunweg:
And disable the 'long lines check' in CI, as it doesn't seem to work. The corresponding mathlib linter finds *many* violations. If desired, it seems easier to just enable that linter instead.


grunweg:
(I'm trying right now how much output the flexible and multi-goal linters produce.)

PR #169, opened 2024-11-11
chore: enable the flexible and multi-goal linters

grunweg:
Both of them could help making proofs more robust. For code to be upstreamed, fixing these is required anyway.


grunweg:
CI passes now.

PR #160, opened 2024-11-05
WIP: progress towards Lemma 8.0.1

grunweg:
This PR is like an incubator/a sand box to work on things as they get ready enough for a PR.

PR #171, opened 2024-11-12
feat: progress towards lemma 8.0.1

grunweg:
Extracted from #160.

### DIFF
--- a/Carleson/HolderVanDerCorput.lean
+++ b/Carleson/HolderVanDerCorput.lean
@@ -15,6 +15,68 @@ open scoped NNReal ENNReal ComplexConjugate
 def cutoff (R t : ℝ) (x y : X) : ℝ≥0 :=
   ⟨max 0 (1 - dist x y / (t * R)), by positivity⟩
 
+variable {R t : ℝ} {x y : X}
+
+lemma cutoff_Lipschitz (hR : 0 < R) (ht : 0 < t) :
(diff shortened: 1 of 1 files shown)


Carleson/HolderVanDerCorput.lean
```diff
@@ -15,6 +15,68 @@ open scoped NNReal ENNReal ComplexConjugate
 def cutoff (R t : ℝ) (x y : X) : ℝ≥0 :=
   ⟨max 0 (1 - dist x y / (t * R)), by positivity⟩
 
+variable {R t : ℝ} {x y : X}
+
+lemma cutoff_Lipschitz (hR : 0 < R) (ht : 0 < t) :
+    LipschitzWith (max 0 ⟨(1 / (t * R)), by positivity⟩) (fun y ↦ cutoff R t x y) := by
```
fpvandoorn, line 21: `max 0 : Real -> NNReal` is usually written [Real.toNNReal](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/NNReal/Defs.html#Real.toNNReal)

grunweg: Thanks! In this case, conceptually one really obtains a maximum, though (so .toNNReal is misleading). Thinking about it: one could just omit the `max` entirely; I'll do that.

PR #175, opened 2024-11-14
chore(Discrete/ForestUnion): getting rid of this

madvorak:


fpvandoorn:
Seems fine. Is there a particular reason you prefer not to use `this`?

madvorak:
> Seems fine. Is there a particular reason you prefer not to use `this`?

It is for when you are in rush to finish your proof. Not something that should stay in the code forever IMO. Most notably, when you get into rush again, you won't be able to use `this` again.

In all projects, I tend to remove `this` during "peaceful periods" (and for every release).

PR #176, opened 2024-11-14
Lemma 7.5.6

lakesare:
[[blueprint]](https://florisvandoorn.com/carleson/blueprint/treesection.html#limited-scale-impact)

This is ready and can be merged!

### TODOs

- [x] prove `have notMax : ¬IsMax J`
- [x] add a typo correction (64⁻¹ version of the hypothesis) to blueprint
- [x] add `\leanok`
- [x] refactor
- [x] use the `have ⟨J'', belongs, plusOne⟩` lemma that I factored-out in another PR
- [x] rename some variables (xxx, yyy)
- [x] fix spacing
- [x] remove "REFACTORED" comments


fpvandoorn:
I'm looking at the proof in the blueprint, and I think there is indeed a gap where you have your sorry. I'll ask the harmonic analysis group here what the right way to fix this is.

fpvandoorn:
Nevermind, the proof is fine.
In your goal, assume by contradiction that `s J + 3 < 𝔰 p`. Then you can prove `¬IsMax J` using `Grid.isMax_iff` and `Grid.le_topCube`

lakesare:
Great, that worked, thanks!

lakesare:
Is it okay if I leave refactoring to a later PR (either by myself or someone else, maybe add `#TODO refactor` to keep track of what theorems need to be refactored) and merge it, or is it better to refactor right away?

Assume I'll do minimal preparation before the merge, e.g. prepend every auxiliary lemma with `limited_scale_impact` (so `lemma first_estimate` would be renamed to `lemma limited_scale_impact__first_estimate`).

fpvandoorn:
I think it's nice if you clean it up yourself first, there's no hurry to get this merged right away.

lakesare:
@fpvandoorn, this is refactored and can be merged!

#### Some notes
- Some lemmas are named nontraditionally, e.g. `IF_subset_THEN_distance_between_centers`, this is just to maximize readability for myself, thought it's ok while they are restricted to this file/while we're making sure we want them as separate lemmas (will be clear as proofs in this file develop). Can rename it mathlib-style if you like!
- All calculational lemmas are named as `calculation_N` (`calculation_1`, `calculation_2`, etc.). This is to indicate that these are just one-off utils.
- Re:Calculations.lean - so far I feel that's a nice way to structure the proofs, factoring calculational lemmas out brought `lemma limited_scale_impact` to readability immediately. I only factored out calculations used in `lemma limited_scale_impact` in this PR to avoid excessive changes.
- I'm touching some other proofs in this PR, but only to remove excessive `↑` arrows, and to reuse the sublemmas I introduced in this PR.

Carleson/ForestOperator/LargeSeparation.lean
```diff
@@ -40,13 +41,38 @@ def holderFunction (f₁ f₂ : X → ℂ)  (J : Grid X) (x : X) : ℂ :=
...
+    dist J p ≥ r := by
+  rw [disjoint_iff_inter_eq_empty, inter_comm] at h
+  by_contra! contr
+  apply (Set.mem_empty_iff_false p).mp
+  rw [← h]
+  apply (Set.mem_inter_iff ..).mpr
+  apply mem_ball_comm.mp at contr
+  exact ⟨belongs, contr⟩
+
+theorem dist_triangle5 (a b c d e : X) :
```
lakesare, line 61:
```suggestion
lemma dist_triangle5 (a b c d e : X) :
```

fpvandoorn:
Oops, I missed that this was ready in the last week. I'll look at it on Monday.

Carleson/ForestOperator/LargeSeparation.lean
```diff
@@ -40,13 +41,38 @@ def holderFunction (f₁ f₂ : X → ℂ)  (J : Grid X) (x : X) : ℂ :=
   χ t u₁ u₂ J x * (exp (.I * 𝒬 u₁ x) * adjointCarlesonSum (t u₁) f₁ x) *
   conj (exp (.I * 𝒬 u₂ x) * adjointCarlesonSum (t u₂ ∩ 𝔖₀ t u₁ u₂) f₂ x)
 
-/-! ### Subsection 7.5.1 and Lemma 7.5.2 -/
+/- AUXILIARY LEMMAS:START -/
+lemma IF_subset_THEN_distance_between_centers (subset : (J : Set X) ⊆ J') :
```
fpvandoorn, line 45: As a remark, the Mathlib naming convention suggests something like this:
```suggestion
lemma dist_c_c_lt_of_subset (subset : (J : Set X) ⊆ J') :
```
or `dist_lt_of_subset` if you want to be more verbose.
It contains roughly the same information. It might be useful to train yourself to get used to names like this, since you'll be using those in Mathlib names.

fpvandoorn (commented):
This looks good to me as-is, so I'll merge it now.

Thanks a lot!

PR #184, opened 2024-11-21
Task 62

pitmonticone:


fpvandoorn:
Thanks! Your intermediate lemma might duplicate some work about Vitali-sets (see `Vitali.exists_disjoint_subfamily_covering_enlargement_ball` in the same file), but I'm not sure if that is exactly true, since that requires that the radii are bounded above (maybe Mathlib has some more general versions that can be used).

PR #193, opened 2024-12-05
Lemma 6 3 2

mariainesdff:


fpvandoorn:
Nice that this proof works now. Thanks!

PR #196, opened 2024-12-23
Suggestion - stop tracking .vscode folder

lakesare:
**Workspace Settings** override **User Settings**, so the only way for the developer to edit these settings is to directly edit `.vscode/settings.json`, which gives excessive file changes in git diff.

Unless we want to enforce some of these vscode settings, I suggest we stop tracking them.


fpvandoorn:
I would like to have some useful default settings in the repository (Mathlib has them, too).
If you want to override these settings, does it help to configure language-specific user settings? Those should override the workspace settings:
https://code.visualstudio.com/docs/getstarted/settings#_settings-precedence

If you want, I can also be a bit less opinionated in these settings, and remove some of the ones that Mathlib also doesn't have.

lakesare:
> Mathlib has them

Oh interesting, looks like we use shared vscode settings to compensate a bit for the lack of linters in Lean

Nope it's fine if these are intentional, I thought it's a copypaste! Let's leave them as they are.

PR #199, opened 2024-12-28
feat: Lemma 5.1.2

sgouezel:
It was harder than I expected :-)

Highlights:
* In the blueprint, the proof starts with "By Lemma 5.4.8, we can decompose U_3 as a disjoint union of at most 4n+13 collections...". I don't know where the 4n+13 comes from (in the end, I got 4n+12). I didn't find an easy argument here and I had to come up with some covering lemma relying on the fact that dyadic cubes are disjoint or included in one another (and it has to play some role, because without this condition there are counterexamples where you can not extract disjoint subfamilies just from a bounded stacking condition) -- see `exists_maximal_disjoint_covering_subfamily` below.
* I refactored `BoundedCompactSupport` (because I needed it for a measure which was not the ambient measure in the `MeasureSpace`). As hinted in todos in the files, I went for the version requiring that the range is bounded, and that the function is strongly measurable (requiring ae strongly measurable would feel weird here since the other conditions are everywhere conditions, not almost everywhere conditions)
* In `ProofData`, I added the assumptions that `F` and `G` are bounded (which is the case in the blueprint)
* I found a better constant than the one of the blueprint (the blueprint loses `2 ^ {3 a ^ 3}`, but with the very same proof one loses only `2 ^ {a ^ 3}`). I stated the version with the better constant, but also the one with the weaker constant to match exactly the statement in the blueprint.
* As usual, I found several lemmas that should be in mathlib. I put them in the `toMathlib` directory, in files with the same names as the files where they should be moved, for easier upstreaming.


Carleson/TileStructure.lean
```diff
@@ -509,3 +515,156 @@ lemma stackSize_real (C : Set (𝔓 X)) (x : X) : (stackSize C x : ℝ) =
...
+      apply Finset.sum_eq_zero (fun p hp ↦ ?_)
+      simp only [Finset.mem_filter, Finset.mem_univ, true_and, not_exists, not_and,
+        indicator_apply_eq_zero, Pi.one_apply, one_ne_zero, imp_false] at hp hx ⊢
+      exact hx _ hp
+    linarith
+
+/-! ### Decomposing a set of tiles into disjoint subfamilies -/
+
+/-- Given any family of tiles, one can extract a maximal disjoint subfamily, covering everything. -/
+lemma exists_maximal_disjoint_covering_subfamily (A : Set (𝔓 X)) :
```
fpvandoorn, line 540: Is `B` given by the maximal elements in `A`? Is `iteratedMaximalSubfamily` the same as https://florisvandoorn.com/carleson/docs/Carleson/ToMathlib/MinLayer.html#Set.maxLayer ?

fpvandoorn:
This looks really good. In response to your points:

* Yes, the structure of the tiles definitely plays a role, and it's fine to just replace the blueprint's `4n+13` by `4n+12`. The blueprint could have been more clear here.
* Thanks for refactoring `BoundedCompactSupport` so that it accepts a measure. The idea for the previous definition was: "it can be bad on a set of measure 0 in its support, but outside it, it should be easy to ensure it is really 0." But maybe we eventually will want `BoundedCompactSupport` (your version) and `AEBoundedCompactSupport` (a.e. equal to a `BoundedCompactSupport`), and the intermediate version is then a bit artificial. Seems fine to change. @roos-j do you agree?
* In `ProofData` there are the fields `F_subset` and `G_subset`, which imply boundedness.
* For constants of the form `2 ^ {c * a ^ 3}`, then feel free to minimize `c`, and then also edit the blueprint with the minimal (integer) `c`. If a better bound makes the constant more complicated, it's probably good to have a separate result, as you nicely did.
* Thanks!

Besides the above points, this looks good to me, and I'm happy to merge it (I didn't do a detailed Mathlib-level review).

Carleson/TileStructure.lean
sgouezel: I don't really know what the order on `A` is. The subtelty is that there can be several `p` with the same `𝓘 p`, and then you should only pick one of them. If the `maxLayer` abstraction can handle this, this would be much better indeed.

sgouezel:
I've updated the blueprint, replacing `4n+13` by `4n+12` and explaining how one get `2 ^ {433 a^3}` instead of `2 ^ {435 a ^ 3}`. I've also removed `isBounded_F` and `isBounded_G` from `PreProofData`, deducing them from `F_subset` and `G_subset` instead. Also, I've moved sum estimates to a new file, as I will also need them for Lemma 5.1.3, and made some cleanup. To me, this looks good to go.

roos-j:
> * Thanks for refactoring `BoundedCompactSupport` so that it accepts a measure. The idea for the previous definition was: "it can be bad on a set of measure 0 in its support, but outside it, it should be easy to ensure it is really 0." But maybe we eventually will want `BoundedCompactSupport` (your version) and `AEBoundedCompactSupport` (a.e. equal to a `BoundedCompactSupport`), and the intermediate version is then a bit artificial. Seems fine to change. @roos-j do you agree?

I like this version of `BoundedCompactSupport`; this is what I had in mind and I believe this is the most helpful variant for this project. Thank you @sgouezel

Carleson/TileStructure.lean
fpvandoorn: Ah, my bad, that indeed doesn't work here.

fpvandoorn:
Sorry for my slow responses so far. If you fix any errors, I'll merge it quickly now. It also looks good to me now.

sgouezel:
There should be conflicts with things already upstreamed. Let me fix those, and I'll let you know when it's ready.

sgouezel:
Should be good.

fpvandoorn:
Thanks!

PR #200, opened 2025-01-01
feat: Lemma 5.1.3

sgouezel:
The proof diverges just a little bit from the blueprint because some already formalized lemmas don't match exactly the blueprint statement. For instance, Lemma 5.5.2 in the blueprint involves n sets, while the Lean version uses n+1 sets (which is slightly weaker, but enough to complete the proof of Lemma 5.1.3). Both versions are true (it would even be true with n / 50 sets). I don't know if you want to reconcile the blueprint and the formalization, by changing one or the other -- or just let it go, and accept local divergences like that which have no consequence whatsoever.


fpvandoorn:
Thanks, this looks great! I hope the computations weren't too painful.

I would like to at the end of the project claim that the proofs in the blueprint are completely correct, since they correspond to the formalization. I don't mind minor deviations too much (e.g. the blueprint indexes from `1` to `n` when Lean indexes from `0` to `n-1`), but please update the blueprint to reflect the changes you made to the proof.
The fact that Lemma 5.1.3 has an off-by-1 error is indeed a small mistake, and I'll try to fix that (or please fix it yourself if you want).

sgouezel:
ok, I'll fix the small mistakes in the other lemmas compared to the blueprint, and adjust my proof accordingly.

sgouezel:
> Thanks, this looks great! I hope the computations weren't too painful.

Not really painful: it's quite relaxing to know that you can't make mistakes, so you just follow your nose and see what it gives you in the end, not trying to be clever (and using `simp` and `ring` to do the computations for you).

I've fixed the Lean statement and proof of Lemma 5.5.2 and adjusted the indexing in the blueprint, so now everything should correspond perfectly.

fpvandoorn:
Thanks again!

PR #204, opened 2025-01-09
7.1.6

js2357:
The lemma `BoundedCompactSupport_approxOnCube` is currently stated for functions with codomain ℂ, even though the more natural setting is a codomain of `E'` with `RCLike E'`, because I ran into technical issues with the generalization and only needed it for ℂ. (EDIT: Actually, for the proof of Lemma 7.2.1 I need it for ℝ also. For now I'll write separate lemmas for ℝ and ℂ, but I'm open to a more elegant solution.)


fpvandoorn:
I don't exactly understand your comment. Replacing one line with
```
lemma BoundedCompactSupport_approxOnCube {𝕜 : Type*} [RCLike 𝕜] (C : Set (Grid X)) (f : X → 𝕜) :
```
works fine for me locally. If you want to generalize it to 
```
lemma BoundedCompactSupport_approxOnCube {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E] 
    (C : Set (Grid X)) (f : X → E) :
    BoundedCompactSupport (approxOnCube C f) :=
```
then we have to generalize the predicate `BoundedCompactSupport` first.

js2357:
That's what I had in mind. I was trying to use `E'` instead of `𝕜` to be consistent with the notation in the definition of `approxOnCube`; I guess one of the other assumptions on `E'` was getting pulled in and causing the problem. Thanks.

fpvandoorn:
But maybe you can generalize from `ℂ` to `RCLike` for now, especially if you need it also for `ℝ`?

js2357:
That's exactly what I planned to do. I'm about to PR Lemma 7.2.1 which includes all the work here (with some minor changes, including the `RCLike` issue), so you can ignore this PR if you want.

js2357:
I just realized that my other comment was a bit vague; I meant that what you did was what I had in mind, not your comment about generalizing further.

fpvandoorn:
Thanks. I'll merge this first, so please merge the other to master in case there is a merge conflict.

PR #205, opened 2025-01-14
Lemma 7.2.1

js2357:
- The constant in the theorem was incorrect. I fixed it in Lean and in the blueprint. There was a comment in the tex file for the blueprint suggesting 2^(255a^3), but I think that was because of a misplaced parenthesis in the proof; it can be brought down to 2^(152a^3) with no extra effort.
- To avoid any confusion, I left a comment in the code mentioning that the constant had been changed. You'll probably want to remove that eventually.
- The value of `CMB` (the constant in `hasStrongType_MB`) is better than what's in the blueprint. For the application in Lemma 7.2.1, the blueprint said it would be 2^(2a+1) but it was actually 2^(a+1.5). I left it as 2^(2a+1) in the blueprint to keep the blueprint internally consistent, because I wasn't prepared to explain where 2^(a+1.5) came from.
- My changes to the blueprint were more substantial than usual, though still relatively mild. Besides the constant and the misplaced parenthesis, I also changed the argument a little. (I'm not sure whether the original proof was valid, but I didn't put too much effort into it because I think my version is clearer anyway.) I also took the liberty of defining a notation to make the proof a bit more readable.
- The statements of `boundary_operator_bound` and `nontangential_operator_bound` both forgot the constants, so I added them. More interestingly, I removed the `toReal` from both of them.
- I added `ENNReal.lintegral_Lp_smul` in ToMathlib. I stated it using integrals rather than using `eLpNorm` because that appears to be the standard that Mathlib follows for theorems about the Lp norm for `ENNReal`-valued functions (e.g., in `Mathlib.MeasureTheory.Integral.MeanInequalities`). But it would work better for me if it were stated using `eLpNorm`, so if Mathlib doesn't have a preference, we can change the statement.
- I proved `boundaryOperator_lt_top`, but then it turned out not to be needed. It's currently commented out in `PointwiseEstimate.lean` in case it becomes relevant later.
- I noticed that `TileStructure.Forest.biUnion_𝓛` doesn't match the statement in the blueprint. I assume the version in the blueprint is a typo and was meant to say what `biUnion_𝓛` says, but I didn't edit it.
- I removed the comment that I had previously left on `Set.Countable.measure_biUnion_le_lintegral` in `HardyLittlewood.lean`, since it's no longer relevant


fpvandoorn:
Not sure how the discrepancy in the blueprint and Lean of `biUnion_𝓛` stayed so long. I just committed a fix to the blueprint. I'll look at the rest later (it might take a few days).

fpvandoorn:
It's only a ~month that `eLpNorm` accepts functions into `ENNReal`, and Mathlib almost exclusively still uses it for functions into normed spaces. Many Mathlib lemmas should be generalized to `ENorm` or similar classes (which should be upstreamed from here once they're ready).

fpvandoorn:
Other than that, this looks great. Thanks!

PR #211, opened 2025-01-17
Lemma 7.6.1, `BoundedCompactSupport.prod_mul`

roos-j:
In this PR:
- some content in helper file `BoundedCompactSupport`, mainly `prod_mul`; meant towards making this file more useful
- proof of Lemma 7.6.1 (very similar to Lemma 7.5.1 by @lakesare)

The proofs of Lemma 7.5.1/7.6.1 could be unified/simplified, but it may not be worth the effort at this point.


fpvandoorn:
Thanks!

PR #212, opened 2025-01-17
Fix definition of PreProofData.hasBoundedStrongType_Tstar

js2357:
The definition of `PreProofData.hasBoundedStrongType_Tstar` said that `(nontangentialOperator K · · |>.toReal)` has bounded strong type, but I think the `toReal` is a mistake; with the `toReal` present, a case where the nontangential operator is ∞ everywhere would qualify as bounded strong type.


fpvandoorn:
Yes, we want to get rid of all the `|>.toReal` in `HasWeakType` and `HasStrongType`. It's a bit tricky to do that all at once without breaking the proofs. But this is helpful!

fpvandoorn:
Related: #210

PR #213, opened 2025-01-19
(WIP) Redefine upperRadius

js2357:
With the current definition of `upperRadius`, it will always be infinity. (Infinity will be in the set that we're taking a supremum of, due to the `toReal` causing infinity to be treated as 0.) This is an attempt to redefine `upperRadius` to avoid that problem, by taking a sup over reals instead. I think a real-valued `upperRadius` will also be nicer to work with, at least once we have some API to deal with the pitfalls of the real-valued sup.

In order for the real-valued sup to behave nicely, we need to know that the set we're taking the sup of is bounded. Our assumptions ensure that this is the case in non-trivial situations, but we might need an additional assumption to rule out trivial cases. Do any of the existing assumptions ensure that d_B is not identically 0?


js2357:
Wrong branch

PR #214, opened 2025-01-19
redefine upperRadius

js2357:
With the current definition of `upperRadius`, it will always be infinity. (Infinity will be in the set that we're taking a supremum of, due to the `toReal` causing infinity to be treated as 0.) This PR redefines `upperRadius` to avoid that problem, by taking a sup over reals instead. I think a real-valued `upperRadius` will also be nicer to work with.

In order for the real-valued sup to behave nicely, we need to know that the set we're taking the sup of is bounded. Our assumptions ensure that this is the case in non-trivial situations, but we might need an additional assumption to rule out trivial cases. Do any of the existing assumptions ensure that d_B is not identically 0?

### DIFF
--- a/Carleson/Defs.lean
+++ b/Carleson/Defs.lean
@@ -193,6 +193,14 @@ def CZOperator (K : X → X → ℂ) (r : ℝ) (f : X → ℂ) (x : X) : ℂ :=
 def upperRadius [FunctionDistances ℝ X] (Q : X → Θ X) (θ : Θ X) (x : X) : ℝ :=
   sSup { r : ℝ | dist_{x, r} θ (Q x) < 1 }
 
+lemma le_upperRadius [FunctionDistances ℝ X] {Q : X → Θ X} {θ : Θ X} {x : X} {r : ℝ}
+    (hr : dist_{x, r} θ (Q x) < 1) : r ≤ upperRadius Q θ x := by
+  have bdd : BddAbove { r : ℝ | dist_{x, r} θ (Q x) < 1 } := by sorry
+  have : { r : ℝ | dist_{x, r} θ (Q x) < 1 }.Nonempty := ⟨r, hr⟩
+  rw [upperRadius, Real.le_sSup_iff bdd this]
+  intro ε hε
+  use r, hr, by linarith
+
 /-- The linearized maximally truncated nontangential Calderon Zygmund operator `T_Q^θ` -/
 def linearizedNontangentialOperator [FunctionDistances ℝ X] (Q : X → Θ X) (θ : Θ X)
     (K : X → X → ℂ) (f : X → ℂ) (x : X) : ℝ≥0∞ :=
--- a/Carleson/Defs.lean
+++ b/Carleson/Defs.lean
@@ -190,14 +190,14 @@ def CZOperator (K : X → X → ℂ) (r : ℝ) (f : X → ℂ) (x : X) : ℂ :=
   ∫ y in {y | dist x y ∈ Ici r}, K x y * f y
 
 /-- `R_Q(θ, x)` defined in (1.0.20). -/
-def upperRadius [FunctionDistances ℝ X] (Q : X → Θ X) (θ : Θ X) (x : X) : ℝ≥0∞ :=
-  sSup { r : ℝ≥0∞ | dist_{x, r.toReal} θ (Q x) < 1 }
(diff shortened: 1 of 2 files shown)


fpvandoorn:
Thanks for spotting this!

My initial thought would be to still have `upperRadius` in `ENNReal`, and define it something like this:

`⨆ (r : ℝ) (h : dist_{x, r} θ (Q x) < 1), ENNReal.ofReal r`

I don't know exactly when `upperRadius` is finite, if `Q` is constantly `θ` I presume it's infinite.
To answer your other question, the following shows that `d_B` is not identically 0:
https://florisvandoorn.com/carleson/docs/Carleson/Defs.html#CompatibleFunctions.localOscillation_le_cdist

fpvandoorn:
I think it can actually happen that `upperRadius` is infinity, and then it's easier to work with an `ENNReal` version. But I'm happy to keep it in Real for now, with a comment "maybe change to `ENNReal` if convenient".

I don't think there is any guarantee that `Θ` is not singleton right now, but we could assume that if convenient. Still: I think there are many cases that `upperRadius` is infinity for some values of 
`θ`.

js2357:
Yeah, I guess we can't work around the case where `θ` = `Q x`. I'll change it back to `ENNReal`.

Carleson/Defs.lean
```diff
@@ -190,14 +190,19 @@ def CZOperator (K : X → X → ℂ) (r : ℝ) (f : X → ℂ) (x : X) : ℂ :=
...
+    (hr : dist_{x, r} θ (Q x) < 1) : r ≤ upperRadius Q θ x := by
+  have bdd : BddAbove { r : ℝ | dist_{x, r} θ (Q x) < 1 } := by sorry
+  exact (Real.le_sSup_iff bdd ⟨r, hr⟩).mpr fun ε hε ↦ ⟨r, hr, add_lt_iff_neg_left.mpr hε⟩
 
 /-- The linearized maximally truncated nontangential Calderon Zygmund operator `T_Q^θ` -/
 def linearizedNontangentialOperator [FunctionDistances ℝ X] (Q : X → Θ X) (θ : Θ X)
     (K : X → X → ℂ) (f : X → ℂ) (x : X) : ℝ≥0∞ :=
   ⨆ (R₁ : ℝ) (x' : X) (_ : dist x x' ≤ R₁),
-  ‖∫ y in {y | ENNReal.ofReal (dist x' y) ∈ Ioo (ENNReal.ofReal R₁) (upperRadius Q θ x')},
+  ‖∫ y in {y | dist x' y ∈ Ioo R₁ (upperRadius Q θ x')},
```
fpvandoorn, line 205: With `upperRadius` in `ENNReal`, I would write this as 
```lean
⨆ (R₁ : ℝ≥0) (x' : X) (_ : dist x x' ≤ R₁),
  ‖∫ y in {y | (nndist x' y : ℝ≥0∞) ∈ Ioo (R₁ : ℝ≥0∞) (upperRadius Q θ x')}, K x' y * f y‖ₑ
```

js2357: 1. I'm not sure we want that. With the original version, the left-hand inequality of the `Ioo` statement has `ENNReal.ofReal` on both sides, so it's relatively easy to translate back to a real-valued inequality. In your new version, one side has an `NNReal` being cast to `ENNReal` and the other side has a real being cast to `ENNReal`. I don't have time to give it much thought right now, but it feels like that would be less convenient.
2. This isn't directly relevant to what I'm currently working on, because I think the reference to `linearizedNontangentialOperator` in the blueprint's proof of Lemma 7.2.2 is a mistake. I think it should be `nontangentialOperator` instead (as appears correctly at the end of the proof).
3. My proof is almost done now (and the heavy lifting is completely finished), so it might make more sense to discuss this after I PR it. It will include a lot of API for sets of the form `{y | ENNReal.ofReal (dist x' y) ∈ Ioo r R}`.

fpvandoorn: Oh oops, I expect that an occurrence of `linearizedNontangentialOperator` in chapter 7 is indeed a mistake. My bad! (See `CONTRIBUTING.md` for the translation of symbols.)
And I'm also not sure if what I suggested is the best option.

I do expect that `ENNReal.ofReal (dist x' y)` is more conventiently written as `edist x' y` (so my `(nndist x' y : ℝ≥0∞)` above should also use `edist`)

PR #216, opened 2025-01-22
WIP: prove first half of Lemma 8.0.1

grunweg:
Some golf and polish is mine, by the main work is done by Sébastien. I'm still going over the diff myself, to see if I want to simplify.

Co-authored by: @sgouezel


grunweg:
As discussed: it could be sensible to review if the use of the Bochner (instead of the Lebesgue) integral is truly necessary.

fpvandoorn:
I guess here using the Bochner integral is reasonable, because it is multiplied by a complex number, so the Lebesgue integral would require another `ENNReal.toReal`.

Thanks to both of you!

PR #221, opened 2025-01-26
Lemma 7.2.2

js2357:
See [here](https://github.com/fpvandoorn/carleson/pull/225) for a version with `Annulus` reworked, expanded, and moved to a separate file.

Proof of Lemma 7.2.2.

- I corrected the blueprint's mistaken reference to `linearizedNontangentialOperator` (which was written correctly as `nontangentialOperator` at the end of the proof).
- I think the blueprint should have used the doubling property 7 + 100a^2 times instead of 6 + 100a^2 times. I changed this in the blueprint; due to the rounding at the end, this change has no effect on the constant in the lemma.
- I improved the constant from 2^(103a^3) to 2^(102a^3) by not rounding prematurely.
- The lemma `CMB_defaultA_two_eq` calculates `CMB (defaultA a) 2`. It's currently in `L2Estimate.lean`, but that's perhaps not the most natural place for it. I think the obvious place would be immediately after the definition of `CMB`, if not for the fact that `defaultA` is currently not imported there.
- I defined `Annulus.oo x r R` to be the set `{y | ENNReal.ofReal (dist x y) ∈ Ioo r R}`, and similarly for `cc` and `ci`. Writing the definition with `ENNReal.ofReal (dist x y)` rather than `edist x y` makes the definition less aesthetically pleasing, but otherwise the `ofReal` version is more convenient for our purposes here.
- I proved a lot of theorems for `Annulus` which turned out to be unnecessary. (In particular, I didn't use `Annulus.oc` and `Annulus.co` at all in the finished proof.) For now, the unnecessary theorems are still there, commented out.


Carleson/Defs.lean
```diff
@@ -271,6 +275,36 @@ lemma MeasureTheory.aestronglyMeasurable_K [IsOneSidedKernel a K] :
 lemma measurable_K_left [IsOneSidedKernel a K] (y : X) : Measurable (K · y) :=
   measurable_K.of_uncurry_right
 
+lemma measurable_K_right [IsOneSidedKernel a K] (x : X) : Measurable (K x) :=
+  measurable_K.of_uncurry_left
+
+lemma integrableOn_K_Icc [IsOpenPosMeasure (volume : Measure X)] [ProperSpace X]
+    [Regular (volume : Measure X)] [IsOneSidedKernel a K] (x : X) (R₁ R₂ : ℝ) (hR₁ : R₁ > 0) :
```
fpvandoorn, line 282: Can you make the non-propositional arguments implicit?

Carleson/ForestOperator/L2Estimate.lean
```diff
@@ -1,18 +1,341 @@
...
-variable {E' : Type*} [NormedAddCommGroup E'] [NormedSpace ℝ E']
+-- Define an annulus in `X` to be a set consisting of all `y` such that `dist x y` lies in an
+-- interval between `r` and `R`. We take `r` and `R` to be in `ENNReal` to accommodate the use
+-- of `upperRadius`.
+namespace Annulus
+
+variable {X : Type*} [MetricSpace X]
+open Set Metric ENNReal
+
+private def oo (x : X) (r R : ℝ≥0∞) := {y | ENNReal.ofReal (dist x y) ∈ Ioo r R}
```
fpvandoorn, line 13: `private` means that these cannot be used outside this file. Maybe use `protected` instead?

fpvandoorn: These might be useful elsewhere (and maybe even upstreamed to Mathlib?). So maybe it's nice to put this in a separate file (for an arbitrary metric space)?

fpvandoorn (commented):
Thanks, this is great! I have a few small comments, but otherwise it looks very good.

Is the reason that `edist` is less convenient just because of missing lemmas about `edist`? Of course, by [edist_dist](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/MetricSpace/Pseudo/Defs.html#edist_dist) it should be almost the same.

Carleson/Defs.lean
js2357: Done.

Carleson/ForestOperator/L2Estimate.lean
js2357: Yes, the only reason to avoid `edist` is to avoid using `edist_dist` every time, since we use `dist` everywhere else. I agree that if we're trying to turn this into Mathlib-ready code, it makes more sense to use `edist`.

If we're going to try to do that, I think it would make sense to settle on a naming convention before expanding the API any more. We would probably (at least eventually) want separate versions using `dist` and `edist`. One option would be similar to what's currently here, something like `Annulus.oo` and `EAnnulus.oo`. I think the other obvious option would be to lean in to the analogy with `Set.Ioo` and do something like `Set.Aoo` and `Set.eAoo`.

fpvandoorn:
Merged #225

PR #229, opened 2025-02-07
WIP: v4.16 Mathlib bump

fpvandoorn:
(creating this PR for Michael Rothgang)

@grunweg please also edit the version in `docbuild/lakefile.toml`.

PR #232, opened 2025-02-10
chore/perf: miscellaneous small tweaks

grunweg:
Some tweaks salvaged from #229, and some small performance improvements I found.


fpvandoorn:
Thanks!
All the comments in `ENorm` will be outdated quickly, since these lemmas will be upstreamed in generalized form, but LGTM.

PR #239, opened 2025-02-19
perf: a few performance improvements

fpvandoorn:
less than I hoped, there was nothing that was exceptionally slow


fpvandoorn:
For future reference, the build time of this PR is actually 8s slower than latest master, but looking at recent builds that is well within the variance. Gonna merge this anyway, since I tested that individual declarations got faster (though we can check if the hacks are obsolete in the next Mathlib bump)

PR #242, opened 2025-02-22
Lemma 7.5.7

Parcly-Taxel:
Pending ironing out of a gap on Zulip.

### DIFF
--- a/Carleson/ToMathlib/Data/ENNReal.lean
+++ b/Carleson/ToMathlib/Data/ENNReal.lean
@@ -1,0 +1,57 @@
+import Mathlib.Topology.Instances.NNReal.Defs
+
+/-! ## `ENNReal` manipulation lemmas -/
+
+open Function Set
+open scoped NNReal
+
+variable {α ι : Type*} {s : Set ι} {t : Finset α}
+
+namespace ENNReal
+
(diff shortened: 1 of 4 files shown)


Carleson/ToMathlib/Data/ENNReal.lean
```diff
@@ -0,0 +1,57 @@
...
+/-! ## `ENNReal` manipulation lemmas -/
+
+open Function Set
+open scoped NNReal
+
+variable {α ι : Type*} {s : Set ι} {t : Finset α}
+
+namespace ENNReal
+
+lemma coe_iSup_NNReal {f : ι → ℝ≥0} (hf : ∃ B, ∀ a, f a ≤ B) :
```
fpvandoorn, line 12: In line with `ENNReal.coe_iSup`, this should probably be called `ENNReal.coe_biSup` (we use `biSup` for supremums where the index is bounded to a set).
Also: `hf` should probably state `BddAbove (Set.range f)` ([bddAbove_def](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Bounds/Basic.html#bddAbove_def) states that these are almost the same)

fpvandoorn (approved):
Thanks! Looks good. One comment.

Carleson/ToMathlib/Data/ENNReal.lean
Parcly-Taxel: Done.

PR #248, opened 2025-02-26
Lemma 7.5.10

lakesare:


lakesare:
Two questions about the theorem statement:

1.  **`MB volume 𝓑 c𝓑 r𝓑 (‖f ·‖) x` vs `MB volume 𝓑 c𝓑 r𝓑 f x`** 
  
    In one of the recent commits, [this line](https://github.com/fpvandoorn/carleson/pull/242/files#diff-336482e1c30213fb45dbf42eac14bf77cd04ba56d69cea6ee44be52c260ed9c4L376) was changed, which breaks the code I had for this proof a bit. 
    Changing `MB volume 𝓑 c𝓑 r𝓑 (‖f ·‖) x` to `MB volume 𝓑 c𝓑 r𝓑 f x` in all statements (`lemma local_tree_control`, `lemma global_tree_control2`) fixes it, should we do it?

2. **Parentheses**

    Here is the blueprint for lemma 7.5.10:
    
    
    
    Here is the original statement of lemma 7.5.10:
    
        ⨆ x ∈ ball (c J) (8 * D ^ s J), ‖adjointCarlesonSum (t u₂ ∩ 𝔖₀ t u₁ u₂) f x‖₊ ≤
        ⨅ x ∈ ball (c J) (8⁻¹ * D ^ s J), ‖adjointCarlesonSum (t u₂) f x‖₊ +
        C7_5_10 a * ⨅ x ∈ J, MB volume 𝓑 c𝓑 r𝓑 (‖f ·‖) x := by
        
     Do I understand it correctly that the following was meant instead:
     
        ⨆ x ∈ ball (c J) (8 * D ^ s J), ‖adjointCarlesonSum (t u₂ ∩ 𝔖₀ t u₁ u₂) f x‖₊ ≤
        (⨅ x ∈ ball (c J) (8⁻¹ * D ^ s J), ‖adjointCarlesonSum (t u₂) f x‖₊) +
        C7_5_10 a * ⨅ x ∈ J, MB volume 𝓑 c𝓑 r𝓑 (‖f ·‖) x

Parcly-Taxel:
> Changing `MB volume 𝓑 c𝓑 r𝓑 (‖f ·‖) x` to `MB volume 𝓑 c𝓑 r𝓑 f x` in all statements (`lemma local_tree_control`, `lemma global_tree_control2`) fixes it, should we do it?

Please do – the expression with the `norm` is equivalent to the expression without. The first function applied to `f` in `MB` is an `enorm`, so the extra `norm is redundant.

For the parenthesising of `sup` and `inf`, please note that their priorities sometimes cause the most intuitive syntax to have incorrect semantics (like `Finset.sum`), so use parentheses when you need them.

fpvandoorn:
Oops, I only see your questions now. Yes, I agree with @Parcly-Taxel about removing the norms, those are redundant, and you are correct about the parenthesization mistake in the statement.

fpvandoorn:
For removing the norms: consider doing that as a separate PR that I can merge quickly, since it is likely to conflict with other PRs

PR #249, opened 2025-02-28
Lemma 7.5.2 and 7.6.3

Parcly-Taxel:
@lakesare used the wrong disjointness assumption in `moderate_scale_change`, so this has been fixed too.


fpvandoorn:
Oops, the misformulation of `moderate_scale_change` was my mistake (although I believe the version I wrote was stronger).

Thanks!

PR #255, opened 2025-03-05
Proofs of Lemmas 7.3.1 and 7.3.3

js2357:
I haven't edited the blueprint yet. You can merge this as-is if you want and I'll make a separate PR for the blueprint changes later, or you can wait and I'll make the blueprint edits here when I have time.

- `C7_3_1_1` and `C7_3_1_2` had to be increased because they depend on Lemma 7.2.1, which had its constant increased from what was originally in the blueprint.
- I defined `C7_3_1_1` to be `2 ^ (202.5 * (a : ℝ) ^ 3)`. I can round up to 203 if you want to stick to integers there for consistency.
- All these constants are formalized as elements of `ℝ≥0`. I haven't changed it yet, but I assume that we want to change these to `ℝ≥0∞`?
- I think the blueprint's version of Lemma 7.3.3 with the constant 2^(200a^3 + 19) is wrong, in the sense that the existing proof can't be trivially repaired to prove it. (In inequality (7.3.8), the 2^(4a) disappears.) But the formalization used the constant 2^(201a^3), which can be proven using the blueprint's method, so I didn't need to change that.
- There was a comment on `density_tree_bound2` predicting that some assumptions on f are superfluous. I don't think either assumption can be easily removed, though the assumption `∀ x, ‖f x‖ ≤ F.indicator 1 x` could easily be weakened to say that the support of `f` is contained in `F`, if we want to do that.

EDIT: I see that the blueprint already got changed. Let me know if you still want me to change it to match what I did.


fpvandoorn:
The formalization looks great! Yes, if you could modify the blueprint to be consistent with the formalization, that would be great!
(I don't care much about the 202.5 or 203, they're both treated as reals anyway.)

js2357:
Blueprint changes are done.

I replaced `measure_ball_le_pow_two'` with `measure_ball_two_le_same_iterate`. While I was doing that, I also noticed that `measure_ball_le_pow_two` and `measure_real_ball_two_le_same_iterate` were the exact same theorem, so I eliminated `measure_ball_le_pow_two`.

fpvandoorn:
Thanks!

PR #257, opened 2025-03-06
Lemma 7.5.9

Parcly-Taxel:


Parcly-Taxel:
> For your second point, can't you apply `le_of_forall_pos_lt_add` and then choose a point in the open ball where you're close enough to the supremum?

@js2357 to what inequality am I to apply that?

fpvandoorn:
(1) Please do increase `𝓑max`. We're not assuming any upper bound on it.

(2) No, we're not assuming the supremum and maxima are actually attained. If `f x <= g y` for all `x` and `y` (in a certain domain) then `sup {f x} <= inf g y`. This is [iSup_le](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/CompleteLattice.html#iSup_le) and [le_iInf](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/CompleteLattice.html#le_iInf) (although for the infimum on the RHS you have to do slightly more work, since you're adding a second term). Does that make sense?

js2357:
> > For your second point, can't you apply `le_of_forall_pos_lt_add` and then choose a point in the open ball where you're close enough to the supremum?
> 
> @js2357 to what inequality am I to apply that?

I didn't have time to go through your proof in detail, but it looks like you're arguing
sup |acs| - inf |acs| = |acs x| - |acs x'| <= bound,
with the problem being that you can't guarantee the existence of an x for which the equality holds. If you choose a close-enough x  (and maybe x') so that
sup |acs| - inf |acs| < |acs x| - |acs x'| +ε <= bound + ε,
I'm assuming the rest of the proof would still work unchanged.

fpvandoorn:
Thanks! Looks good.

fpvandoorn:
One late comment: maybe you can split
```lean
lemma biSup_enorm_ne_top_iff {f : ι → E} :
    ⨆ z ∈ s, ‖f z‖ₑ ≠ ⊤ ↔ IsBounded (f '' s) := by
  sorry
```
out of the proof of `exists_biSup_le_enorm_add_eps`?

PR #258, opened 2025-03-10
chore: bump mathlib

grunweg:
The biggest highlight is the upstreaming of most remaining `ENorm`-related classes. Apart from that, almost no changes.


fpvandoorn:
Thanks!

PR #259, opened 2025-03-10
chore: port one ennnorm theorem to enorm

grunweg:
The new proof can be simplified to avoid some back-and-forth, but compiles.

I found this in an old branch of mine: should we simply land this, or clean it up further first?

### DIFF
--- a/Carleson/ToMathlib/MeasureTheory/Integral/SetIntegral.lean
+++ b/Carleson/ToMathlib/MeasureTheory/Integral/SetIntegral.lean
@@ -4,17 +4,20 @@ open MeasureTheory
 
 /- Put after `setIntegral_re_add_im`, but probably refactor first to use `enorm` -/
 
+-- TODO: give a nice name!
+lemma aux {α 𝕜 : Type*} [RCLike 𝕜] {f : α → 𝕜} (x : α) :
+    starRingEnd 𝕜 (f x / ‖f x‖) * f x = ‖f x‖ := by
+  simp only [div_eq_inv_mul, map_mul, map_inv₀, RCLike.conj_ofReal, mul_assoc, RCLike.conj_mul]
+  norm_cast
+  rcases eq_or_ne (‖f x‖) 0 with hx | hx
+  · simp [hx]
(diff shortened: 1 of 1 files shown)


Carleson/ToMathlib/MeasureTheory/Integral/SetIntegral.lean
```diff
@@ -24,3 +24,29 @@ lemma ennnorm_integral_starRingEnd_mul_eq_lintegral_ennnorm
   conv_rhs => rw [← this]
   simp only [Real.norm_eq_abs, Real.toNNReal_abs]
   rfl
+
+-- move to Mathlib.Analysis.Normed.Module.Basic, next to nnnorm_algebraMap'
+lemma enorm_algebraMap' {𝕜 : Type*} (𝕜' : Type*) [NormedField 𝕜] [SeminormedRing 𝕜'] [NormedAlgebra 𝕜 𝕜']
+    [NormOneClass 𝕜'] (x : 𝕜) : ‖(algebraMap 𝕜 𝕜') x‖ₑ = ‖x‖ₑ := by
+  simp_rw [enorm_eq_nnnorm]; rw [nnnorm_algebraMap']
+
+lemma enorm_integral_starRingEnd_mul_eq_lintegral_enorm
```
fpvandoorn, line 33: This proof is not Mathlib-ready: hypothesis `A` has to be a separate lemma, and then we should add the lemma `‖∫ x, ‖f x‖ ∂μ‖ₑ = ∫⁻ x, ‖f x‖ₑ ∂μ`.

fpvandoorn: I'm happy to merge things here that are not Mathlib-ready, but if nothing depends on this, let's clean it up before merging.

grunweg: For the record, everything you say equally holds about the proof I modified - but I agree with both of your points.

grunweg: #301 extracts the said lemma.

PR #261, opened 2025-03-13
Draft PR for Lemma 10.1.2

JasperMS:
This draft PR contains a first stab at lemma 10.1.2 from the blueprint. There are still some sorries left.

I will follow up on Zulip to discuss more; link to thread will follow.

This is my first PR trying to contribute to a Lean project, so bear with me while I figure things out. Please feel free to liberally improve/correct/suggest/explain/whatever. I'm here to learn, and while this lemma has been a journey, it should only mark the beginning.

Remaining sorries:
- [ ] L122 `czoperator_welldefined`: I am not sure if this is already proven somewhere
- [ ] L396 `Measurable (vol x)`: Again may already be proven (though `vol` was very recently added in favour of `Real.vol`)
- [ ] L404 `rw_dom`: Some set work on proving a ball is a union of annuli
- [ ] L535 `ContinuousSMul ℝ≥0∞ ℝ≥0∞`: I feel this should be an oversight that is easily fixed somewhere
- [ ] L570 `Summable`: It's a geometric series, but I didn't know how to approach. May be able to put this in `ℝ≥0∞` to use `ENNReal.summable`

I have been reluctant to change other files since I am new to Lean and associated etiquette. Here are some other suggestions/ideas/observations:
- The use of ENorm, `vol`, and constants in `ℝ≥0` made my life much easier so I am happy with that.
- L342 `laverage_le_globalMaximalFunction` worked well, but it uses `‖ _ ‖₊` instead of `‖ _ ‖ₑ`
- I used `mul_le_mul` a lot but it seems that in NNReal/ENNReal it can be simplified to avoid `case c0 | b0` (e.g. L349 is typical)
- It was a pain to prove measurability of the integrands, although it was a 'simple' composition
- L132 I made `setIntegral_union_2` because I struggled a lot to split an integral using `setIntegral_union` because of the IntegrableOn conditions on the parts instead of the whole
- Am I adding too many intermediate steps into the context? Does this affect performance?
- I am not 100% sure where the assumptions on `g` are actually used (and which). Is it for `czoperator_welldefined`?

Overall I have the feeling, based on the amount of work and the number of intermediate steps needed, that it might be wise to have a discussion on approach (e.g. which general lemmas) for chapter 10 before distributing the work further to individual contributors. And a word of caution that the blueprintś

PR #263, opened 2025-03-14
Refactor to BoundedFiniteSupport

JasperMS:
Split off from #261 referred in [this comment](https://github.com/fpvandoorn/carleson/pull/261#issuecomment-2724425055).

Replaces `(hg : eLpNorm g ∞ < ∞)` with `(hg : MemLp g ∞ volume)` everywhere.

In doing so I stumbled upon `def BoundedFiniteSupport` [here](https://github.com/JasperMS/carleson/commit/ec65a6948205804406caa43d9b5e63adf452f180#diff-df9329a1f992660a7d5c94ea4bee97cd3245d5971297dfc1e2a217f2ee79e1a4L78).

This is the same like many assumptions in chapter 10 (typically `hmg hg h2g`). Should the definition be moved somewhere more generic and applied everywhere? This would reflect the repetitive phrasing from the blueprint:

> _bounded measurable function $g$ on $X$ supported on a set of finite measure_

If so, where should it go? Please advise.


fpvandoorn:
`MemLp` implies AEMeasurable (via AEStronglyMeasurable), so you can remove the separate `Measurable` assumptions.

I guess `BoundedFiniteSupport` can indeed be put in a more basic file (I wasn't if we want to use this generally, or just for that one lemma). Please put it in a new file importing `BoundedCompactSupport` (in the same directory as that file).

Carleson/Classical/HilbertStrongType.lean
```diff
@@ -53,16 +53,16 @@ lemma mean_zero_oscillation {n : ℤ} (hn : n ≠ 0) :
 Note: might not be used if we can use `spectral_projection_bound_lp` below.
 -/
 lemma partial_sum_projection {f : ℝ → ℂ} {n : ℕ}
-    (hmf : Measurable f) (hf : eLpNorm f ∞ < ∞) (periodic_f : f.Periodic (2 * π)) {x : ℝ} :
+    (hmf : Measurable f) (hf : MemLp f ∞ volume) (periodic_f : f.Periodic (2 * π)) {x : ℝ} :
```
fpvandoorn, line 56:
```suggestion
    (hf : MemLp f ∞ volume) (periodic_f : f.Periodic (2 * π)) {x : ℝ} :
```
also below

JasperMS: Fixed. I think it is now ready to merge.

fpvandoorn:
Thanks!

PR #264, opened 2025-03-17
chore(WeakType): generalise to enorm classes, part 1

grunweg:
- Some results were already stated about enorms, but still put in the NormedSpace section: move them to clarify.
- Add one enorm counterpart of a lemma, and use it to generalise two more lemmas.

### DIFF
--- a/Carleson/ToMathlib/WeakType.lean
+++ b/Carleson/ToMathlib/WeakType.lean
@@ -566,22 +566,20 @@ end ContinuousENorm
 
 section NormedGroup
 
--- todo: generalize various results to ENorm.
-
 variable {f g : α → ε}
 section
 variable [TopologicalSpace ε] [ContinuousENorm ε]
 
-lemma distribution_eq_nnnorm {f : α → E} : distribution f t μ =  μ { x | t < ‖f x‖₊ } := rfl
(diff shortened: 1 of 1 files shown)


Carleson/ToMathlib/WeakType.lean
```diff
@@ -560,21 +554,43 @@ lemma distribution_mono (h₁ : ∀ᵐ x ∂μ, ‖f x‖ₑ ≤ ‖g x‖ₑ) (
...
+
+end ContinuousENorm
+
+section NormedGroup
+
+variable {f g : α → ε}
+section
+variable [TopologicalSpace ε] [ContinuousENorm ε]
+
+-- TODO: generalise this to the ENorm context: requires generalising the HasSMul context for •
```
fpvandoorn, line 583: This is not generalizable to `ENorm`, which is fine. However, a variant where you take a scalar multiplication w.r.t. `NNReal` is generalizable using `ENormedSpace`, and is probably useful (and probably sufficient for all applications).

grunweg:
```suggestion
-- TODO: add an analogue for the ENorm context, using scalar multiplication w.r.t. `NNReal` on an `ENormedSpace`

PR #273, opened 2025-03-22
Some Interval results

JasperMS:
After discussion [on Zulip](https://leanprover.zulipchat.com/#narrow/channel/442935-Carleson/topic/Lemma.2010.2E1.2E2.20and.20Chapter.2010/near/506398730).
Will be used to fix a `sorry` in PR #261.

Before putting more work in proving Annulus versions, a first review of the Interval stuff would be appreciated to avoid rework later.

In particular I am not sure about the imports, and the proofs feel a bit long and clunky to me.

### DIFF
--- a/Carleson/ToMathlib/Interval.lean
+++ b/Carleson/ToMathlib/Interval.lean
@@ -1,9 +1,5 @@
--- I have no idea how to determine the correct imports...
-import Mathlib.Order.Interval.Set.Defs
-import Mathlib.Order.Bounds.Basic
-import Mathlib.Order.SetNotation
 import Mathlib.Data.Set.Lattice
-import Mathlib.Tactic.WLOG
+import Mathlib.Tactic.Common
 
 open Function Order Set
 
@@ -15,15 +11,17 @@ section Preorder
 
 variable [Preorder α]
 
-theorem Ico_subset_Ici (h : c ≤ a): Ico a b ⊆ Ici c := by
-  trans Ico c b
-  . exact Ico_subset_Ico_left h
-  . exact Ico_subset_Ici_self
+theorem Ico_subset_Ici (h : c ≤ a): Ico a b ⊆ Ici c :=
+  (Ico_subset_Ico_left h).trans Ico_subset_Ici_self
 
-theorem Ioc_subset_Ioi (h : c ≤ a): Ioc a b ⊆ Ioi c := by
-  trans Ioc c b
-  . exact Ioc_subset_Ioc_left h
-  . exact Ioc_subset_Ioi_self
+theorem Icc_subset_Ici (h : c ≤ a): Icc a b ⊆ Ici c :=
+  (Icc_subset_Icc_left h).trans Icc_subset_Ici_self
+
+theorem Ioc_subset_Ioi (h : c ≤ a): Ioc a b ⊆ Ioi c :=
+  (Ioc_subset_Ioc_left h).trans Ioc_subset_Ioi_self
+
+theorem Ioo_subset_Ioi (h : c ≤ a): Ioo a b ⊆ Ioi c :=
+  (Ioo_subset_Ioo_left h).trans Ioo_subset_Ioi_self
 
 end Preorder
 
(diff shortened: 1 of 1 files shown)


Carleson/ToMathlib/Interval.lean
```diff
@@ -0,0 +1,134 @@
...
+      let this := hcontra i
+      rw [mem_Ioc, not_and, not_le] at this
+      exact this hind
+
+/-
+Pairwise disjoint statements to above results. It also holds for Ioo; is that valuable?
+-/
+theorem pairwise_disjoint_Ico_monotone {f : ℕ → α} (hf : Monotone f) :
+    Pairwise (Function.onFun Disjoint fun (i : ℕ) => Ico (f i) (f (i+1))) := by
+  simp (config := { unfoldPartialApp := true }) only [Function.onFun]
```
fpvandoorn, line 102:
```suggestion
  simp +unfoldPartialApp only [Function.onFun]
```
(this is a pretty new syntax for changing configurations)

Note: `unfold Function.onFun` also works.

Carleson/ToMathlib/Interval.lean
```diff
@@ -0,0 +1,134 @@
...
+  simp_rw [Set.disjoint_iff]
+  intro i j hinej
+  wlog hij : i < j generalizing i j
+  . rw [not_lt] at hij
+    have := @this j i (Ne.symm hinej) (lt_of_le_of_ne hij (Ne.symm hinej))
```
fpvandoorn, line 107:
```suggestion
    have := this hinej.symm (hij.lt_of_ne hinej.symm)
```

Carleson/ToMathlib/Interval.lean
```diff
@@ -0,0 +1,134 @@
...
+    rw [inter_comm] at this
+    exact this
```
fpvandoorn, line 109:
```suggestion
    rwa [inter_comm]
```
(or `rwa [inter_comm] at this` also work)

Carleson/ToMathlib/Interval.lean
```diff
@@ -0,0 +1,134 @@
...
+    intro i
+    induction i
+    case zero => exact ha
+    case succ i hind =>
+      let this := hcontra i
+      rw [mem_Ico, not_and, not_lt] at this
+      exact this hind
+
+theorem iUnion_Ioc_eq_Ioi {f : ℕ → α} (hf : Monotone f) (h2f : ¬BddAbove (range f)) :
+    ⋃ (i : Nat), Ioc (f i) (f (i+1)) = Ioi (f 0) := by
```
fpvandoorn, line 68: It's unfortunate that the proof here is almost the same. Not sure how to reuse the previous result.

Carleson/ToMathlib/Interval.lean
```diff
@@ -0,0 +1,134 @@
...
+  . exact Ioc_subset_Ioc_left h
+  . exact Ioc_subset_Ioi_self
+
+end Preorder
+
+section LinearOrder
+
+variable [LinearOrder α]
+
+/- Statement can be optimised (or a variant) to also work if α has a maximum and f attains it (in particular : α is finite)
```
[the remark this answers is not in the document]
JasperMS: After thinking more about it, I found the more general statement to be false. E.g. $f$ constant. You miss the last bit of the interval. In fact, from the theorem result you can prove that $f$ is not bounded. I therefore kept it at "not bounded above".

Carleson/ToMathlib/Interval.lean
```diff
@@ -0,0 +1,134 @@
...
+    rw [not_exists] at hcontra
+    suffices BddAbove (range f) by
+      contradiction
+    rw [bddAbove_def]
+    use a
+    suffices ∀ i, f i ≤ a by
+      intro y hy
+      obtain ⟨i, hi⟩ := hy
+      rw [← hi]
+      exact this i
```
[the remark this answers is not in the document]
JasperMS: Unfortunately this doesn't work with the other proof, but at least here it saves some lines.

Carleson/ToMathlib/Interval.lean
JasperMS: It feels tedious but I also don't see an immediate way to avoid it.

Carleson/ToMathlib/Interval.lean
```diff
@@ -0,0 +1,134 @@
...
+    case succ i hind =>
+      let this := hcontra i
+      rw [mem_Ioc, not_and, not_le] at this
+      exact this hind
+
+/-
+Pairwise disjoint statements to above results. It also holds for Ioo; is that valuable?
+-/
+theorem pairwise_disjoint_Ico_monotone {f : ℕ → α} (hf : Monotone f) :
+    Pairwise (Function.onFun Disjoint fun (i : ℕ) => Ico (f i) (f (i+1))) := by
```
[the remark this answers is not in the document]
JasperMS: I had to give $\iota$ a linear order (otherwise disjoint parts might overlap under $f$).
If you make $f$ antitone, then the interval endpoints should be swapped, so that is at least not the same statement as we have here.

Carleson/ToMathlib/Interval.lean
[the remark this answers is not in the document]
fpvandoorn: Good point! I guess the weaker condition does work for the `Ioc` version?

PR #276, opened 2025-03-27
Apply Interval results to Annulus

JasperMS:
Straightforward continuation of #273 applied to Annulus. With this, it's ready to resolve a sorry in #261.


fpvandoorn:
Thanks!

PR #277, opened 2025-03-28
chore: cleanup

pitmonticone:


Carleson/ToMathlib/Data/ENNReal.lean
```diff
@@ -54,7 +55,7 @@ lemma finsetSum_biSup {f : α → ι → ℝ≥0∞}
   induction t using Finset.cons_induction with
   | empty => simp
   | cons a t ha ihs =>
-    simp_rw [Finset.sum_cons, ihs]
+    simp only [Finset.sum_cons, ihs]
```
fpvandoorn, line 58: this is a regression (also below)

fpvandoorn (commented):
I'm a bit worried about the performance of aesop, which is sometimes slow because it calls `simp_all` first. But if it's not noticeably slow in the places you've changed, I'm happy to merge this.

Carleson/ToMathlib/Data/ENNReal.lean
pitmonticone: Ops, you're right. Sorry.

pitmonticone: Done

pitmonticone:
No, `aesop` does not seem to degrade performance significantly in the places where I introduced it.

PR #283, opened 2025-04-04
feat(WeakType): generalise lemmas to enorm

grunweg:
Generalises most of the remaining file to enorm: one section about the layercake formula remains (and will be changed in a future PR). This requires generalising a handful of mathlib lemmas: leanprover-community/mathlib4#23707 contains these lemmas and upstreams some changes for this PR. (Other lemmas need `ENormedSpace` to be in mathlib first.)

To be discussed: is the new generality of the const_{s}mul lemmas appropriate: the new version covers e.g. enormed spaces, but does *not* cover normed spaces with a k-action which does not induce an R-action: e.g., real or complex normed spaces are fine, normed spaces over a finite field (or the p-adics?) are not.


Carleson/ToMathlib/ENorm.lean
```diff
@@ -74,4 +88,82 @@ lemma esub_zero [TopologicalSpace E] [ENormedAddCommSubMonoid E] {x : E} : x - 0
...
+    by_cases ht : t = 0
+    · simp [ht]
+      fun_prop
+    have : Continuous fun (x : ε) ↦ ‖t • x‖ₑ := by
+      simp_rw [ENormedSpace.enorm_smul]
+      fun_prop
+    -- careful: ε need not be a metric space
+    -- preimage of an open set U ⊆ ε is precisely t⁻¹ ⬝ U => suffices to show this map is open
+    -- which it is, I presume? haven't thought it through
+    sorry
```
grunweg, line 124: To be discussed: is this instance true in general? Do we even need it? (Assuming `ContinuousConstSMul NNReal \eps` should certainly be fine in practice, and clearly satisfied in all settings we want.)

grunweg: I just pushed a change, instead assuming that instance in the lemmas that need it.

PR #289, opened 2025-04-05
feat: Proposition 2.0.5

sgouezel:
Also adapt the statement, the statement of Lemma 8.0.1 and the blueprint, assuming that the function is Hölder on `B(z, 2R)` instead of just `B (z, R)` as initially claimed: as discussed on Zulip, this is not enough to obtain the statements.


fpvandoorn:
Thanks, this is great!

PR #287, opened 2025-04-05
chore: bump to `v4.19.0-rc2`

pitmonticone:
This PR bumps the toolchain to `v4.19.0-rc2`.

Further work is required to fix `Carleson/TileStructure.lean`. Help is very much appreciated with this one @grunweg.

PR #293, opened 2025-04-12
chore: bump mathlib to 4.19.0rc2

sgouezel:
Builds on #287 (but I couldn't push to this branch because I work from a fork, hence the new PR).

The issue that #287 couldn't solve is solved here in an ugly way (declaring an instance by hand with `letI`). Probably related to the new structure elaboration. I'll report the difficulty on Zulip.

The file `MeasureReal` has been upstreamed, and is therefore deleted in this PR. During the upstreaming, there was the comment that the `nnreal` version had no API, and was probably not a good idea anyway, so it has been deleted. This means that, in the project, I have replaced the lines `volume.nnreal s` with `(volume s).toNNReal`. In most cases, these are an indication that one should probably work with ENNReal instead (just like we switched from NNNorm to ENorm), but I haven't made any such change in the PR, which is a pure bump PR without any other change.


Carleson/ToMathlib/RealInterpolation.lean
```diff
@@ -1227,13 +1227,14 @@ lemma lintegral_add_right_Ioi (f : ℝ → ENNReal) {a b : ℝ} :
 
 lemma lintegral_scale_constant (f: ℝ → ENNReal) {a : ℝ} (h : a ≠ 0):
     ∫⁻ x : ℝ, f (a*x) = ENNReal.ofReal |a⁻¹| * ∫⁻ x, f x := by
+  change ∫⁻ x : ℝ, f (a*x) = ENNReal.ofReal |a⁻¹| • ∫⁻ x, f x
```
grunweg, line 1230: This looks identical to the goal to me. Am I overlooking something/why is this line needed?

Carleson/ToMathlib/RealInterpolation.lean
```diff
@@ -1227,13 +1227,14 @@ lemma lintegral_add_right_Ioi (f : ℝ → ENNReal) {a b : ℝ} :
...
   rw [← @lintegral_smul_measure, MeasurePreserving.lintegral_comp_emb]
   · exact measure_preserving_scaling h
   · exact measurableEmbedding_mulLeft₀ h
 
-lemma lintegral_scale_constant_preimage (f: ℝ → ENNReal) {a : ℝ} (h : a ≠ 0)
-    {s : Set ℝ}:
+lemma lintegral_scale_constant_preimage (f: ℝ → ENNReal) {a : ℝ} (h : a ≠ 0) {s : Set ℝ} :
     ∫⁻ x : ℝ in (fun z : ℝ ↦ a * z)⁻¹' s, f (a*x) = ENNReal.ofReal |a⁻¹| * ∫⁻ x : ℝ in s, f x := by
+  change ∫⁻ x : ℝ in (fun z : ℝ ↦ a * z)⁻¹' s, f (a*x) = ENNReal.ofReal |a⁻¹| • ∫⁻ x : ℝ in s, f x
```
grunweg, line 1237: Idem

Carleson/Classical/CarlesonOnTheRealLine.lean
```diff
@@ -369,6 +369,7 @@ instance compatibleFunctions_R : CompatibleFunctions ℝ ℝ (2 ^ 4) where
   eq_zero := by
     use 0
     intro f
+    change f 0 = 0
```
grunweg, line 372: Do you know why this is needed?

grunweg (commented):
Three small questions: otherwise, looks good to me - thanks for doing this!

Carleson/ToMathlib/RealInterpolation.lean
sgouezel: A multiplication has been replaced by an smul.

Carleson/Classical/CarlesonOnTheRealLine.lean
sgouezel: My guess is that, due to the change of structure elaboration, a coercion is not in place at some point while it used to be, so the shape of the goal has changed (it used to be with the coercion, now it's without it, but still these are definitionally the same goals)

Carleson/ToMathlib/RealInterpolation.lean
grunweg: Would you like to change this to e.g. `rw [smul_def]` (or something like that)? Optional.

sgouezel: Done.

grunweg:
Thanks! I'll merge this when CI has finished.

pitmonticone:
Thank you so much @sgouezel!

fpvandoorn:
Thanks! Removing the `nnreal` version sounds good to me. I think moving away from `NNReal` is usually a good thing.

PR #300, opened 2025-04-15
feat(WeakType): generalise most remaining lemmas to enorm

grunweg:
One section about the layercake formula remains (and will be changed in a future PR). This requires generalising a handful of mathlib lemmas: https://github.com/leanprover-community/mathlib4/pull/23707 contained some of these, others need ENormedSpace to be in mathlib first.

For most of this file, there is a trade-off about generality to be made: currently, there is no typeclass unifying both a normed space over a field and an ENormedSpace. Hence, we usually add enorm variants of existing lemmas and keep the existing versions.
We make one change, though: the normed space becomes primed; the enorm version stays unprimed.

### DIFF
--- a/Carleson/ToMathlib/WeakType.lean
+++ b/Carleson/ToMathlib/WeakType.lean
@@ -21,7 +21,7 @@ variable {α 𝕜 E : Type*} {m : MeasurableSpace α}
 lemma ENNNorm_absolute_homogeneous {c : 𝕜} (z : E) : ofNNReal ‖c • z‖₊ = ↑‖c‖₊ * ↑‖z‖₊ :=
   (toReal_eq_toReal_iff' coe_ne_top coe_ne_top).mp (norm_smul c z)
 
-lemma enorm_absolute_homogeneous {c : 𝕜} (z : E) : ‖c • z‖ₑ = ‖c‖ₑ * ‖z‖ₑ :=
+lemma enorm_absolute_homogeneous' {c : 𝕜} (z : E) : ‖c • z‖ₑ = ‖c‖ₑ * ‖z‖ₑ :=
   (toReal_eq_toReal_iff' coe_ne_top coe_ne_top).mp (norm_smul c z)
 
 lemma ENNNorm_add_le (y z : E) : ofNNReal ‖y + z‖₊ ≤ ↑‖y‖₊ + ↑‖z‖₊ :=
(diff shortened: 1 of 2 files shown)


Carleson/ToMathlib/WeakType.lean
```diff
@@ -571,43 +578,54 @@ variable {f g : α → ε}
...
-  refine fun f hf ↦ ⟨AEStronglyMeasurable.const_smul (h f hf).1 k, eLpNorm_const_smul_le.trans ?_⟩
+  refine fun f hf ↦ ⟨AEStronglyMeasurable.const_smul (h f hf).1 k, eLpNorm_const_smul_le'.trans ?_⟩
   simp only [ENNReal.smul_def, smul_eq_mul, coe_mul, mul_assoc]
   gcongr
   exact (h f hf).2
 
-lemma HasStrongType.const_mul {E' α α' : Type*} [NormedRing E']
-    {_x : MeasurableSpace α} {_x' : MeasurableSpace α'} {T : (α → ε) → (α' → E')} {p p' : ℝ≥0∞}
-    {μ : Measure α} {ν : Measure α'} {c : ℝ≥0∞} (h : HasStrongType T p p' μ ν c) (e : E') :
+-- XXX: is this the statement we want?
```
fpvandoorn, line 618: No, this statement is near-useless. We won't have any `T` of that type.

I guess for now just add a version where `T` has codomain `ENNReal`

If we want a common generalization, we might need a `ENormedCommSemiring` class, which I guess is on the way towards `ENormedSemifield`.

fpvandoorn: Don't forget the primed version.

fpvandoorn (commented):
LGTM

grunweg:
Note to self: before merging, land #301 and fix the build in HardyLittlewood.

Carleson/ToMathlib/WeakType.lean
grunweg: Ah, the first observation helps indeed, changed. I wasn't sure if we wanted primed version of all results - but I can certainly add them.

grunweg: Added the primed versions of everything else.

grunweg:
Update: before merging, it could be useful to look at the diamond in HardyLittlewood.

fpvandoorn:
LGTM now

fpvandoorn:
There was no diamond. You assumed an unnecessarily strong class (`ENormedSpace`) on ε for lemmas in WeakType, which didn't exist in the application, where we only 
have an `NormedAddCommGroup` and no scalar multiplication.

I recommend avoiding `convert` if you get a lot of subgoals.

grunweg:
Ah, thanks for clearing that up. Now everything makes sense!
And I'll make sure to heed the `convert` advice in the future.

PR #304, opened 2025-04-22
chore: bump mathlib

pitmonticone:


grunweg:
Thanks, LGTM. Can you also remove the `weak` prefix from the `linter.flexible` option in the lakefile? With recent mathlib, that should no longer be necessary.

fpvandoorn:
Thanks

PR #315, opened 2025-04-27
Task 118 update

jim-portegies:
This is meant to conclude Task 118: removing the assumption on the finiteness of balls for the strongtype and weaktype estimates on the maximal function. It builds on top of #294 but it also refactors some code to avoid code duplication. I make this a separate PR because in principle the other one was already ready for merge.

Please note that I have adapted to constant in `hasWeakType_maximalFunction`, and I would not know how to circumvent this. So far, it did not essentially affect the rest of the files.


fpvandoorn:
Thanks!

PR #316, opened 2025-04-27
Lemma 10.1.3

ldiedering:
Proves Lemma 10.1.3 up to `czoperator_welldefined` which is yet to be incorporated from the pull request for Lemma 10.1.2. 
Might need some more cleanup, I'm happy to take suggestions.


fpvandoorn:
Thanks! Looks good.

One comment for next time: there is a soft line limit of 100 characters per line (VSCode should also draw a line in the 100th column. It's nice if you could add line-breaks so that you don't go over that limit.

PR #325, opened 2025-05-05
Lemma 10.1.4 (both parts)

JasperMS:
Update:

* Now also includes the F2 estimate
* I think `HasBoundedWeakType` should be refactored to use `BoundedFiniteSupport` instead of three separate conditions. If desired I can make a separate PR for that after this lands.

The below has already been addressed.

<strike>Some points of attention:

* This lemma will need `globalMaximalFunction_lt_top` from `HardyLittlewood.lean : 710` but don't know how to do that. Marked with two sorries at `L707`, `L710`
* To prove `czOperator_aemeasurable` (or similar) at `L671` I was stuck with dealing with functions defined as integrals. Probably a matter of me not knowing the right Mathlib results. Or is this where `hT` is actually put to use?
* In applying some `indicator` results (`L720`), I encountered "deterministic timeout" errors, with Lean slowing down and eating significant memory/CPU. Was fixed by an extra `intro y`.
* Should the result `AEMeasurable.restrict` be tagged with `fun_prop`?</strike>

### DIFF
--- a/Carleson/ToMathlib/HardyLittlewood.lean
+++ b/Carleson/ToMathlib/HardyLittlewood.lean
@@ -677,6 +677,13 @@ theorem hasWeakType_maximalFunction_equal_exponents
 def C_weakType_maximalFunction (A p₁ p₂ : ℝ≥0) :=
   if p₁ = p₂ then (ofNNReal A) ^ (2 / p₁ : ℝ) else C2_0_6 A p₁ p₂
 
+lemma C_weakType_maximalFunction_lt_top {A p₁ p₂ : ℝ≥0} :
+    C_weakType_maximalFunction A p₁ p₂ < ∞ := by
+  unfold C_weakType_maximalFunction
+  split_ifs with hps
+  · apply rpow_lt_top_of_nonneg (by positivity) (by simp)
+  · simp
+
 /-- `hasStrongType_maximalFunction` minus the assumption `hR`, but where `p₁ = p₂` is possible and
 we only conclude a weak-type estimate. -/
 theorem hasWeakType_maximalFunction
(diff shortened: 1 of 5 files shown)


fpvandoorn:
> Some points of attention:
> 
>     * This lemma will need `globalMaximalFunction_lt_top` from `HardyLittlewood.lean : 710` but don't know how to do that. Marked with two sorries at `L707`, `L710`

Combine `Bornology.IsBounded.eLpNorm_top_lt_top` with `MB_le_eLpNormEssSup` (note that `eLpNorm f ⊤ μ` is `eLpNormEssSup f μ` by definition) and then `maximalFunction_eq_MB` to use it. Hopefully that helps, otherwise let me know. (feel free to formulate the lemma `globalMaximalFunction_lt_top`, it might be more convenient with the assumption `MemLp u ⊤ μ`)

>     * To prove `czOperator_aemeasurable` (or similar) at `L671` I was stuck with dealing with functions defined as integrals. Probably a matter of me not knowing the right Mathlib results. Or is this where `hT` is actually put to use?

You'll want [MeasureTheory.AEStronglyMeasurable.integral_prod_right'](https://leanprover-community.github.io/mathlib4_docs/Mathlib/MeasureTheory/Integral/Prod.html#MeasureTheory.AEStronglyMeasurable.integral_prod_right'), `hT` should not be needed there. It is possible that `hT` isn't needed for this lemma, but only for later lemmas in this section.

>     * In applying some `indicator` results (`L720`), I encountered "deterministic timeout" errors, with Lean slowing down and eating significant memory/CPU. Was fixed by an extra `intro y`.

While the deterministic timeout is not good, the `apply` tactic is expected to fail. If your goal is `\forall x : A, P x` and you apply a lemma `forall (y : B) (z : C), Q y z`, then Lean tries to see whether `A =?= C` and that `P x =?= Q ?y x`. In this case, your universal quantifier was an unrelated quantification, and this is not expected to work without an `intro`. Lean apparently tries really hard to make `apply` work by unfolding a lot of definitions, which is a bit silly in this case.

>     * Should the result `AEMeasurable.restrict` be tagged with `fun_prop`?

Yes! Please add it to some others in `ToMathlib.Misc`.

PR #329, opened 2025-05-05
refactor(RealInterpolation/Misc): refactor handling of truncations

grunweg:
Refactor the definition of truncations in preparation of generalising everything to `ENormedAddMonoid`s. More specifically,
- change all norms to enorms in that file
  (When generalising this to functions with codomain in an enormed monoid, this is required.)
- make `trunc` and `truncCompl` take a truncation parameter in `ENNReal`. (This follows from the previous step; in practice, this parameter was always non-negative anyway.)
- adapt the definition of `res` and `res'` accordingly: they are still subsets of R, but take an ENNReal truncation parameter. (The value \infty must be handled specially; `toReal` has the wrong behaviour.)
- make the scale choice exponent `d` an ENNReal
- tweak the definition of ToneCouple: its domain and inverse functions should map from ENNReal, hence into ENNReal also. This means the conditions on the range actually become superfluous.
- make the interpolated exponents also take a parameter `t : ENNReal`: overall, this simplifies the file quite a bit.
- Tries to fixes the proofs that break. Some are sorried for future clean-up.

Some notable changes to proofs include
- `linarith` can no longer be used, as `ENNReal` is not a linearly ordered commutative ring (but only a semiring)
- working with ENNReal means a few lemmas grow side conditions about positivity or finiteness. The finiteness tactic is quite helpful with the latter.
- a number of proofs actually simplify, as we can avoid back-and-forth between ENNReal and Real \o/
- Golf some proofs a bit, using e.g. `gcongr with x`, `filter_upwards` or `simp_all`

This does not fix all proofs: there are about 55 sorries left (of varying difficulty). Those can be fixed in follow-up PRs.

This PR is inspired by and largely overlaps with #295: once this file compiles, we can close the other one.

Co-authored-by: @jim-portegies


grunweg:
Status update: everything compiles without errors now :tada: (The bad news: this is because I sorried many broken proofs; there are about 40 sorries remaining. Some are relatively straightforward, others might require deeper fixes.)

The most recent one is mathematically interesting: in `Subadditive_trunc_from_SubadditiveOn_Lp₀p₁`, at truncation parameter \infty the result is not always true. It is true if I assume the constant A is at least 1. (That will cause the constant in the real interpolation theorem to also grow.) Is that an issue?

Alternatively, does something go wrong (mathematically) if we weaken `Subadditive_trunc` to not demand anything for t=\infty?

grunweg:
Potential next steps:
- make one clean file with the better exponent estimates: I presume that using t as ENNReal everywhere is what we actually want
(Redo all estimates there, then remove the old file and rename lemmas back.)
- think about the integration domains again: perhaps some of these actually need an ENNReal...

fpvandoorn:
Growing the interpolation constant is absolutely no issue. Assuming `A \ge 1` or having constants that involve `max A 1` is also absolutely no issue. In most sensible cases you will be able to prove `A \ge 1`.
`Subadditive_trunc` is trivial when `t`/`a` is infinity (assuming `A \ge 1`), but I expect it will be nice to assume it also in that case, because that will make it easier when you use it.

PR #336, opened 2025-05-19
Lemma 10.1.5

JasperMS:
Proof was straightforward.

I added `czoperator_sub` directly because it felt like the extra effort to state and prove linearity of `czoperator` was not worth it.

### DIFF
--- a/Carleson/TwoSidedCarleson/NontangentialOperator.lean
+++ b/Carleson/TwoSidedCarleson/NontangentialOperator.lean
@@ -847,13 +847,9 @@ theorem cotlar_estimate (ha : 4 ≤ a)
       exact this
     exact not_mem_subset subset_union_right ((mem_compl_iff _ _).mp hx'.1)
   apply add_le_add (add_le_add h1x' h2x' |> enorm_sub_le.trans) (by rfl) |> le_trans
-  rw [add_assoc, C10_1_3_def, C10_1_4_def, C10_1_5_def]
-  gcongr
-  rw [← add_mul]
-  gcongr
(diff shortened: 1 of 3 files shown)


Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -798,15 +798,63 @@ theorem cotlar_set_F₂ (ha : 4 ≤ a) (hr : 0 < r) (hR : r ≤ R) [IsOneSidedKe
...
+    exact not_mem_subset subset_union_right ((mem_compl_iff _ _).mp hx'.1)
+  apply add_le_add (add_le_add h1x' h2x' |> enorm_sub_le.trans) (by rfl) |> le_trans
+  rw [add_assoc, C10_1_3_def, C10_1_4_def, C10_1_5_def]
+  gcongr
+  rw [← add_mul]
+  gcongr
+  push_cast
+  nth_rw 5 [pow_succ]
+  rw [mul_two]
+  gcongr <;> simp
```
fpvandoorn, line 857:
```suggestion
  rw [add_assoc, C10_1_3_def, C10_1_4_def, C10_1_5_def, ← add_mul]
  conv_rhs => rw [pow_succ, mul_two]
  push_cast
  gcongr <;> simp
```
here a slightly reordered version that is a bit shorter

fpvandoorn (commented):
Thanks, looks good!

I'm happy that this one was straightforward.

Can you add the appropriate `\leanok` command in the blueprint (at the start of the corresponding proofs in `main.tex`)? (Please also do this for any previous PRs where I forgot to do it myself.)

fpvandoorn:
Thanks!

The `\proves` command is not necessary if you're proving the preceding lemma, so I removed them again.

PR #345, opened 2025-05-20
feat: add a few more lemmas to the finiteness tactic

grunweg:
Split out from #329.


grunweg:
Carleson has built locally - let me go ahead and merge it.

PR #351, opened 2025-05-26
Lower-semicontinuity of maximal function - Task 135

jim-portegies:
Proof of the lower-semicontinuity of the maximal function and lower-semicontinuity of the global maximal function.

Although not used in the proof, I also proved `continuous_average_ball` under an extra assumption. Without that assumption I proved `lowerSemicontinuous_integral_ball`. I'm not sure these are still useful, because they are currently not used. Feel free to discard.


fpvandoorn:
Thanks! I think they could be nice results for Mathlib eventually, but I don't think we'll use them in Carleson.

PR #353, opened 2025-05-27
chore: bump mathlib

pitmonticone:


pitmonticone:
I believe it would be preferable—for the sake of maintaining a clean and accessible tag and release history—to first fix the remaining error (for which I’m unsure of the best approach; suggestions or direct pushes are very welcome). Then, we can bump to include [leanprover-community/mathlib4#25111](https://github.com/leanprover-community/mathlib4/pull/25111) in a separate PR.

grunweg:
I pushed a fix to WeakType. I'm not sure why this is required now: the only change to [the relevant mathlib file](https://github.com/leanprover-community/mathlib4/blame/master/Mathlib/Analysis/Normed/MulAction.lean) could be [this one](https://github.com/leanprover-community/mathlib4/commit/3c616a3f07070b2159cc01acf52fd137c1840fc8), which adds a shortcut instance for normed smul - and also changes the hypotheses demanded for `enorm_smul_eq`.

I could look into the mathlib file, if this is strictly required. On the other hand, the relevant lemmas in WeakType are not actually used in carleson, so merely adding a TODO to investigate is also fine. What do you think?

CC @fpvandoorn in case you have opinions

grunweg:
I pushed a more targeted fix; it seems CI passes with this. :tada:

PR #355, opened 2025-05-27
First part of Lemma 7.6.2, to Equation (7.6.3)

Parcly-Taxel:
Note that the blueprint is already slightly incorrect in this part of the proof – the splitting by scale difference must be done before applying the inequalities, not the other way around.


fpvandoorn (commented):
Thanks, this is great!

PR #356, opened 2025-05-28
feat(Forests): tweak and estimate constant C7_4_4

grunweg:
Prove the main estimate of constants necessary for Lemma 7.4.4: this involves a bit of by hand estimation. Since that is necessary anyway, improve the constant slightly (and update the blueprint to match).
Extracted from #354.


fpvandoorn:
LGTM, feel free to merge already.

PR #362, opened 2025-05-30
WIP: Lemma 10.1.6

JasperMS:
Proof itself is done, two technicalities have me bested:

- [ ] (L851) Measurability of $$T_*^r g$$ (how to pass through the indexed suprema?)
- [x] (L917) The formalisation of the constant 2.0.6' of the global maximal function

Quite some `fun_prop`s were missing, that took me a while to figure out.

@grunweg FYI, I encountered some stuff about `eLpNorm` which is not yet formulated in ENorm classes (L859-865).

Also, please note `enorm_NNReal`. If I tag it with `simp`, stuff breaks (maybe/probably also outside of Carleson). But I think it should have simp. Should that be a separate PR?


fpvandoorn:
L851: This is unfortunately tricky, we're missing some Mathlib lemmas. Ideally we want to know that `fun y ↦ ⨆ x ∈ s, f x y` is measurable if `s` is an `OrdConnected` subset of `Real` (or `NNReal`/`ENNReal`), `f x` is measurable for each `x`, and `f` has some continuity condition (hopefully left- or right-continuous is enough). 
A very similar proof is given in `carlesonOperatorReal_measurable` (and maybe also another time in Carleson)
Perhaps Mathlib's `measurable_of_tendsto_metrizable'` is also useful.

L917: Perhaps you can move `CMB_defaultA_two_eq` to `HardyLittlewood`. That should also help in your case.

`enorm_NNReal` seems like a fine simp-lemma. A separate PR might be useful for that, if you have to fix other Carleson-files for that.

Note: `eLpNorm_const_smul'` is not true as currently written, you're not assuming enough properties. Is this sufficient for you? https://florisvandoorn.com/carleson/docs/Carleson/ToMathlib/ENorm.html#ENormedSpace.enorm_smul_eq_smul

JasperMS:
Constant and eLpNorm lemma have been fixed. I think that can be all for this PR.

I don't feel confident about the `sorry` for the measurability since you mentioned it can probably be done more efficiently at appropriate generality. Maybe this can be a new task?

Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -848,22 +848,90 @@ theorem cotlar_estimate (ha : 4 ≤ a)
   push_cast
   gcongr <;> simp
 
-/-- The constant used in `simple_nontangential_operator`. -/
+lemma simpleNontangentialOperator_aestronglyMeasurable {g : X → ℂ} (hg : BoundedFiniteSupport g):
```
fpvandoorn, line 852:
```suggestion
-- c.f. discussion in https://github.com/fpvandoorn/carleson/pull/362
lemma simpleNontangentialOperator_aestronglyMeasurable {g : X → ℂ} (hg : BoundedFiniteSupport g):
```

PR #364, opened 2025-05-31
Lemma 7.4.5

Parcly-Taxel:
See [Zulip](https://leanprover.zulipchat.com/#narrow/channel/442935-Carleson/topic/Small.20imprecision.20in.20Lemma.208.2E0.2E1/near/521415072) for a potentially fatal error in the proof that I found.

### DIFF
--- a/blueprint/src/chapter/main.tex
+++ b/blueprint/src/chapter/main.tex
@@ -5838,7 +5838,7 @@ $$
     $$
     Using \Cref{Holder-van-der-Corput} with the ball $B(J)$, we bound this by
     $$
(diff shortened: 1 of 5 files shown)


blueprint/src/chapter/main.tex
```diff
@@ -5555,6 +5555,9 @@ \subsection{H\"older estimates for adjoint tree operators}
         B(J) := B(c(J), 8D^{s(J)}).
     \end{equation*}
     We also denote
+    \begin{equation*}
+     BB(J) := B(c(J), 16D^{s(J)}),
```
fpvandoorn:
```suggestion
     B'(J) := B(c(J), 16D^{s(J)}),
```
On paper it's quite uncommon to write operators/variables with more than one letter. Maybe you can prime the `B(J)` notation?

Parcly-Taxel: This PR's ready now, including the above change and fixing `not_mem` deprecations.

PR #372, opened 2025-06-02
chore: complete proof of Lemma 7.4.4

grunweg:
Also fix three deprecation warnings, and golf a lemma in GridStructure using a helper lemma I added.


grunweg:
Thanks!

PR #374, opened 2025-06-03
Lemma 7.7.3

Parcly-Taxel:
Includes a golfing of Lemma 7.4.4.


Carleson/ForestOperator/Forests.lean
```diff
@@ -405,6 +323,11 @@ lemma pairwiseDisjoint_rowDecomp :
 
 @[simp] lemma rowDecomp_apply : t.rowDecomp j u = t u := rfl
 
+open scoped Classical in
+/-- The definition of `T_{ℜ_j}*f(x)`, defined above Lemma 7.7.2. -/
+def rowCarlesonSum (t : Forest X n) (j : ℕ) (f : X → ℂ) (x : X) : ℂ :=
```
fpvandoorn:
```suggestion
def adjointCarlesonRowSum (t : Forest X n) (j : ℕ) (f : X → ℂ) (x : X) : ℂ :=
```
(or something else that has `adjoint` in the name, otherwise it becomes very confusing).

Otherwise, LGTM.

Parcly-Taxel: Done. 👌

PR #375, opened 2025-06-03
chore: bump mathlib

grunweg:
Do not merge yet: this should wait for a doc-gen release (and perhaps me adding a few missing enorm lemmas). Let's merge this at the end of the week.

Also untested if this requires any further adaptations (perhaps not).
Possible follow-up: use mathlib's linter set (or at least investigate doing so).


fpvandoorn:
Thanks for not immediately merging. I indeed appreciate waiting a bit so that issues [like this](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/.E2.9C.94.20Lean.20server.20not.20starting.20in.20VSCode/with/522109006) are ironed out. Merging at the end of the week, if there are no other problems with the release candidate, sounds good to me.

PR #378, opened 2025-06-04
feat: use mathlib's new linter set

grunweg:
- Use all of mathlib's linters --- except for the long line linter, which we disable (as it was before this PR).
- Fix all warnings for the `missingEnd` linter: the errors tell you already what to do, this is easy to fix and seems harmless to do.
- The flexible linter is now imported in Mathlib.Init, so removing the `weak` prefix seems safe.


grunweg:
Looking at the build logs: I'm happy to disable the line length linter, for instance.

grunweg:
Local build is green: let me merge this right away.

PR #379, opened 2025-06-04
The forest operator proposition

Parcly-Taxel:
In the last commit I moved the Carleson and adjoint operators `carlesonOn, carlesonSum, adjointCarleson, adjointCarlesonSum` to a common file that can be used by both the `ForestOperator` and `Antichain` folders, since lemmas on both sides use the operators in their statements.

PR #380, opened 2025-06-06
Cleanup of Lemma 6.1.5 proof + natural powers

Parcly-Taxel:
The first commit fills in the last sorry in Lemma 6.1.5's proof. To facilitate the cleanup, in the second commit I changed some exponents appearing in constants to be `ℕ`-valued rather than `ℝ`- or `ℝ≥0`-valued, where such a change was possible:
```lean
/-- The constant appearing in part 3 of Lemma 2.1.3. -/
def D2_1_3 (a : ℕ) : ℝ≥0 := 2 ^ (150 * a ^ 3)
```
Overall this leads to fewer extra assumptions that need to be proven and fewer type conversion steps.

### DIFF
--- a/Carleson/Antichain/TileCorrelation.lean
+++ b/Carleson/Antichain/TileCorrelation.lean
@@ -176,10 +176,6 @@ lemma correlation_kernel_bound {s₁ s₂ : ℤ} {x₁ x₂ : X} (hs : s₁ ≤ 
     _ ≤ _ := by
       rw [← ENNReal.add_div]; refine ENNReal.div_le_div_right ?_ _
       rw [C2_1_3, C_6_2_1]; norm_cast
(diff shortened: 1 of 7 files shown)


fpvandoorn:
This looks nice! Can you also add the `\leanok` for Lemma 6.1.5?

fpvandoorn:
> > This looks nice! Can you also add the `\leanok` for Lemma 6.1.5?
> 
> Done. Though I am still waiting on what you and your team have to say on the logical error I found while doing #379, which is the last missing step I need (see [here](https://leanprover.zulipchat.com/#narrow/channel/442935-Carleson/topic/Problems.20in.20the.20forest.20operator.20proposition/near/522496771)).

I am aware. I'm in conversation with Lars about it, and let you know as soon as I know the right answer.

PR #381, opened 2025-06-06
feat: if f is in L^p and L^r, it is in L^q for p \leq q \leq r < \infty

grunweg:


ldiedering:
> Sorry for the duplicate work, Leo!

No problem, I hadn't worked on it for too long. Should I now try to finish the proof (is that the meaning of the awaiting-author label) or are you @grunweg still working on it?

ldiedering:
> Since you asked about finishing it: yes, that would be very welcome. I'm happy to take a look at it then.

Ok. I had just removed that part because it looked like there is very little left anyway, but the of course I'm even more happy to do it :)

grunweg:
I see you made great progress here. (Also, thanks a lot for generalising a few of these WeakType lemmas and the real interpolation sorries you fixed.) In principle, I could just merge this once this passes CI (the errors look rather minor) --- but since all of this is in `ToMathlib`, I wonder whether a more careful approach is useful.

For instance, the changes to `WeakType`, fixing the proof of `rpow_le_rpow_of_exponent_le_base_le_enorm` or the lemma generalisations in that file could each be their own PRs --- or I could try to review those with closer-to-mathlib scrutiny. (I can make time for that in the next days for sure.) Is polishing these changes something you're interested in?
It's fine if not; you can also merge this and I'll find time to make any clean-ups later. (I could also push these to this branch directly: if so, let's make sure to avoid conflicts.)

Please let me know what you think!

grunweg:
(Generalising the typeclasses in RealInterpolation was on my list --- after fixing most of the sorries in that directory. But if they can be done now, why not?) Sure, let's merge now (once CI passes) --- and we can polish later.

ldiedering:
> (Generalising the typeclasses in RealInterpolation was on my list --- after fixing most of the sorries in that directory. But if they can be done now, why not?) 

I mean, I hope the remaining sorries are still true after changing the typeclasses. But if not, the proof would still need to be restructured when changing typeclasses later so it's probably best to do it now. Anyway, from my experience until now, most things will just continue to work as before (after minor changes). 

> Sure, let's merge now (once CI passes) --- and we can polish later.

It finally builds now after I fixed more files. But at least to me it looks like these changes made sense since they allowed to weaken some more hypotheses and even fix the last sorry in the `HardyLittlewood.lean` file. 
I'll still have to fix some linters now...

grunweg:
And the build both passes and lints :tada:

grunweg:
CI passes; I just updated the PR description.

Thanks again for your big refactoring in passing. My worry with doing so now was that fixing the sorries in RealInterpolation would involves two entangled issues (the transition to t being an ENNReal and the enorm refactoring) --- but I'm sure all of them are still provable, and happy to sort this out later. So let's merge this.

PR #382, opened 2025-06-08
Prettify `Antichain`

Parcly-Taxel:
This PR does not add any new content and has no relation to #379. Rather, it
* prettifies `AntichainOperator.lean` (and fixes a couple of copy-paste errors)
* applies the same treatment as in #380 to `Antichain` constants, namely `ℕ`-valued exponents
* removes underscores in constant names for consistency with the `ForestOperator` constants.


fpvandoorn:
Looks good! Thanks!

PR #386, opened 2025-06-11
Fix definition of GeneralCase (partial progress on Lemma 10.2.5)

js2357:
- The definition of `GeneralCase` had the wrong inequality. This is now fixed.
- The following two lemmas about `GeneralCase` are proven using the correct definition. 
- The assumption `(hf : BoundedFiniteSupport f)` turned out to be unnecessary for the proof of `isOpen_MB_preimage_Ioi`, and thus unnecessary for a lot of subsequent work. I removed the assumption from any definitions/lemmas that were already completed without using it, but left it as an assumption for any lemma that doesn't have a proof yet.

### DIFF
--- a/Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
+++ b/Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
@@ -215,6 +215,7 @@ lemma iUnion_czPartition (ha : 4 ≤ a) {hX : GeneralCase f α} :
     simp_all
 
 open scoped Classical in
+variable (f) in
 /-- The function `g` in Lemma 10.2.5. (both cases) -/
 def czApproximation (ha : 4 ≤ a) (α : ℝ≥0∞) (x : X) : ℂ :=
   if hX : GeneralCase f α then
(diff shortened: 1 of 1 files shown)


Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
```diff
@@ -153,153 +165,153 @@ lemma czBall_subset_czPartition (ha : 4 ≤ a) {hf : BoundedFiniteSupport f} {hX
...
+    (hX : ¬ GeneralCase f α) : czApproximation (f := f) ha α x = ⨍ y, f y := by
   simp [czApproximation, hX]
 
 /-- The function `b_i` in Lemma 10.2.5 (general case). -/
-def czRemainder' (ha : 4 ≤ a)(hf : BoundedFiniteSupport f) (hX : GeneralCase f α) (i : ℕ) (x : X) : ℂ :=
-  (czPartition ha hf hX i).indicator (f - czApproximation ha hf α) x
+def czRemainder' (ha : 4 ≤ a) (hX : GeneralCase f α) (i : ℕ) (x : X) : ℂ :=
+  (czPartition ha hX i).indicator (f - czApproximation (f := f) ha α) x
 
 /-- The function `b = ∑ⱼ bⱼ` introduced below Lemma 10.2.5.
```
fpvandoorn, line 246:
```suggestion
variable (f) in
/-- The function `b = ∑ⱼ bⱼ` introduced below Lemma 10.2.5.
```

Can you make `f` explicit in all *definitions* that don't depend on an assumption on `f`?

fpvandoorn (commented):
Thanks! One comments, otherwise looks good.

Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
js2357: Done

PR #387, opened 2025-06-12
Proofs of tsum_czRemainder' and measurable_czApproximation (Lemma 10.2.5)

js2357:
The proofs are complete, aside from one sorry in `measurable_czApproximation` for the measurability of `f` (which is supposed to be an assumption, according to the blueprint). The comment on `BoundedFiniteSupport` says that it should include measurability, but I figured I would check in before changing that, in case some other change was intended.

### DIFF
--- a/Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
+++ b/Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
@@ -291,12 +291,9 @@ lemma aemeasurable_czApproximation (ha : 4 ≤ a) {hf : AEMeasurable f} :
                  have hx : ∃ j, x ∈ czPartition ha hX j := ⟨i, hxi⟩
                  simpa [czA, hx, czPartition_pairwiseDisjoint' ha hx.choose_spec hxi] using hi
   rw [this]
-  apply MeasurableSet.union
-  · have := Measurable.exists (MeasurableSet.czPartition ha hX · |>.mem)
-    measurability
-  · refine MeasurableSet.sUnion ((to_countable _).image _) ?_
-    rintro _ ⟨i, ⟨_, rfl⟩⟩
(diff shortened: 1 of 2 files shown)


Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
```diff
@@ -250,14 +257,44 @@ def czRemainder (ha : 4 ≤ a) (α : ℝ≥0∞) (x : X) : ℂ :=
...
+      · have ⟨i, hi⟩ := hxi
+        refine Or.inr ⟨czPartition ha hX hxi.choose, ⟨mem_image_of_mem _ ?_, hxi.choose_spec⟩⟩
+        simpa [czApproximation, hX, hxi] using hx
+      · exact Or.inl ⟨hxi, by simpa [czApproximation, hxi, hX] using hx⟩
+    · cases hx with
+      | inl hx => simpa [czApproximation, hX, mem_setOf_eq ▸ mem_setOf_eq ▸ hx.1] using hx.2
+      | inr hx => obtain ⟨b, ⟨⟨i, ⟨hiM, rfl⟩⟩, hxb⟩⟩ := hx; simpa [czApproximation_def_of_mem ha hxb]
+  rw [this]
+  apply MeasurableSet.union
+  · have : Measurable f := by sorry
```
fpvandoorn: `BoundedFiniteSupport` gives you `AEStronglyMeasurable`, which gives `AEMeasurable`. There is indeed a mismatch here.

Maybe you could refomulate the statement and the name of this lemma to use `AEMeasurable`?

js2357: I did that. I also changed the assumption on `f` to `AEMeasurable f` since that's the only part of `BoundedFiniteSupport` that's actually needed.

PR #389, opened 2025-06-13
Rationalise `ProofData`

Parcly-Taxel:
* `volume_F_pos` is totally unused and `volume_G_pos` only appears in two calc steps which are still true when `volume G = 0`. They can probably be removed.
* The blueprint says that `S` is the smallest integer such that the ranges of `σ₁, σ₂` are contained in `[-S, S]` **and** `F, G` are contained in the ball `B(o, D^S / 4)`. Yet `defaultS` as currently defined does not take into account the latter property. We fix this, after the aforementioned provisional removal of `volume_FG_pos`, by adding `IsBounded F` and `IsBounded G` to `PreProofData` and **unifying** `ProofData` into it, so that only `ProofData` remains. The ball inclusion can be proven from `IsBounded F` and `IsBounded G`.


fpvandoorn:
Thanks!

PR #393, opened 2025-06-16
Lemma 10.2.4

Parcly-Taxel:
The hypothesis `4 ≤ a` is not needed for `ball_covering`, so it has been removed from that lemma and the proved dependents. On the other hand, we perform a large case distinction depending on whether the point set `U` generated by Zorn is finite or infinite.


grunweg:
Thanks for your PR! This will conflict with #395, which completes the proof of Lemma 10.2.4: do you mind if that one gets merged first?

Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
```diff
@@ -99,70 +409,69 @@ lemma isOpen_MB_preimage_Ioi (hX : GeneralCase f α) :
...
-abbrev czBall7 (ha : 4 ≤ a) (hX : GeneralCase f α) (i : ℕ) : Set X :=
-  ball (czCenter ha hX i) (7 * czRadius ha hX i)
+abbrev czBall7 (hX : GeneralCase f α) (i : ℕ) : Set X :=
+  ball (czCenter hX i) (7 * czRadius hX i)
 
-lemma czBall_pairwiseDisjoint (ha : 4 ≤ a) {hX : GeneralCase f α} :
-    univ.PairwiseDisjoint fun i ↦ closedBall (czCenter ha hX i) (czRadius ha hX i) :=
-  ball_covering ha (isOpen_MB_preimage_Ioi hX) |>.choose_spec.choose_spec.1
+lemma czBall_pairwiseDisjoint {hX : GeneralCase f α} :
+    univ.PairwiseDisjoint fun i ↦ ball (czCenter hX i) (czRadius hX i) :=
```
grunweg, line 436: Can you say a word about you need to change this? Will this change complicate things downstream?

grunweg:
You have a fair point about the dependency, particularly since this strengthens hypotheses slightly.

> Look at all the unnecessary hypotheses I removed!

That's precisely the part I'm worried about: how do you know the hypothesis is unused if you haven't formalised the proof? I'd rather leave that condition alone in this PR, and land a follow-up to remove it after #395 has been landed.

js2357:
I don't think that the proofs of Lemmas 10.2.5 (#395) and 10.2.6 (#401) ever use `ha` except via `ball_covering`, so it should be safe to remove those assumptions there. I'm also fine with leaving some or all of the `ha`'s in for now and removing them as part of my PRs, if that's more convenient.

Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
Parcly-Taxel: The blueprint never says anything about closed balls in Lemma 10.2.4, only plain and open balls.

grunweg:
@js2357 Thanks for clarifying! In this case, let's just merge this PR first.

PR #396, opened 2025-06-17
Lemma 10.1.8

JasperMS:
It turned out to be easier to first do 10.1.8 to know the exact statements for 10.1.7 needed for the continuity argument to follow through (I used `ContinuousWithinAt.closure_le`).
This will hopefully help to formalize 10.1.7 in a nicer way than to follow the epsilon-delta argument in the blueprint directly.

I didn't add the `leanok` yet because the statements of the `small_annulus` lemmata will still change and that will slightly affect the proof.


Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -943,24 +943,98 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
 
 /-- Part of Lemma 10.1.7, reformulated. -/
 theorem small_annulus_right (ha : 4 ≤ a)
-    (hT : ∀ r > 0, HasBoundedStrongType (czOperator K r) 2 2 volume volume (C_Ts a))
```
grunweg, line 946: Are you deliberately removing this? Do you know already you won't need this?
(I haven't looked at the math proof deeply, which is why I'm asking. Perhaps this is obvious enough.)

Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -943,24 +943,98 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
...
     nontangentialOperator K f x =
-    ⨆ (R₁ : ℝ) (R₂ : ℝ) (_ : R₁ < R₂) (x' : X) (_ : dist x x' ≤ R₁),
+    ⨆ (R₁ : ℝ) (_: 0 < R₁) (R₂ : ℝ) (_ : R₁ < R₂) (x' : X) (_ : dist x x' < R₁),
     ‖∫ y in ball x' R₂ \ ball x' R₁, K x' y * f y‖ₑ := by
-  sorry
+  let sup : ℝ≥0∞ := ⨆ (R₁ : ℝ) (_: 0 < R₁) (R₂ : ℝ) (_ : R₁ < R₂) (x' : X) (_ : dist x x' < R₁),
+    ‖∫ y in ball x' R₂ \ ball x' R₁, K x' y * f y‖ₑ
+  unfold nontangentialOperator
+  apply le_antisymm
+  all_goals (
```
grunweg, line 965: I have the glaring feeling this can be golfed (perhaps using the `apply_rules` tactic).
But since we're not aiming for mathlib quality in this part of the project, this could probably be merged as-is.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -943,24 +943,98 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
...
+    rw [iSup_le_iff]; intro x'
+    rw [iSup_le_iff]; intro hx'
+  )
+  · have : ∀ (R' : ℝ), R' ∈ Ioo R₁ R₂ → ‖∫ (y : X) in Annulus.oo x' R₁ R₂, K x' y * f y‖ₑ ≤
+        ‖∫ (y : X) in Annulus.oo x' R₁ R', K x' y * f y‖ₑ + sup := by
+      intro R' hR'
+      have : Annulus.oo x' R₁ R₂ = Annulus.oo x' R₁ R' ∪ Annulus.co x' R' R₂ := by
+        exact Annulus.oo_union_co hR'.1 hR'.2.le |>.symm
+      rw [this, setIntegral_union_2 ?dj (by measurability) ?int]
+      case dj => exact disjoint_left.mpr <| fun x hx hx2 ↦ not_lt.mpr hx2.1 hx.2
```
grunweg: I would personally write ?_ ?_ in the refine below, and discharge this side goal with just a focusing dot. That seems a bit more common in mathlib. All of this is a matter of taste, though.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -943,24 +943,98 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
...
+      case int =>
+        simp_rw [← this]
+        apply IntegrableOn.mono_set <| czoperator_welldefined hf hR₁ x'
+        rw [← Annulus.ci_eq]
+        apply Annulus.oo_subset_ci (by rfl)
+      apply le_trans <| enorm_add_le _ _
+      gcongr
+      rw [Annulus.co_eq, inter_comm, ← diff_eq_compl_inter]
+      apply le_trans ?_ <| le_iSup _ (i := R')
```
grunweg, line 986: *Probably* this can be golfed. There ought to be a way, at least (and if it's impossible right now, there should be a better tactic for it). I'd need to check locally to see how, though.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -943,24 +943,98 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
...
+      apply le_trans ?_ <| le_iSup _ (i := R₂)
+      apply le_trans ?_ <| le_iSup _ (i := hR'.2)
+      apply le_trans ?_ <| le_iSup _ (i := x')
+      apply le_trans (by rfl) <| le_iSup _ (i := hx'.trans hR'.1)
+    -- apply continuity
+    have le_R1 : ‖∫ (y : X) in Annulus.oo x' R₁ R₂, K x' y * f y‖ₑ ≤
+        ‖∫ (y : X) in Annulus.oo x' R₁ R₁, K x' y * f y‖ₑ + sup := by
+      refine ContinuousWithinAt.closure_le ?_ ?_ ?_ this
+      · simp [closure_Ioo hR₂.ne, hR₂.le]
+      · apply continuousWithinAt_const
```
grunweg, line 997: You could also inline a short proof term such as this (this is to some extend a matter of taste).

Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -943,24 +943,98 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
...
+      · apply ContinuousWithinAt.add ?_ continuousWithinAt_const
```
grunweg, line 998: Probably (untested), you could also write this as
```
have := small_annulus_right ha hf
fun_prop

File: Carleson/TwoSidedCarleson/NontangentialOperator.lean [hunk 7]
```diff
@@ -943,24 +943,98 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
...
+    simpa using le_R1
+  · have : ∀ (R' : ℝ), R' ∈ Ioo (dist x x') R₁ → ‖∫ (y : X) in ball x' R₂ \ ball x' R₁, K x' y * f y‖ₑ ≤
+        ‖∫ (y : X) in Annulus.oo x' R' R₁, K x' y * f y‖ₑ + nontangentialOperator K f x := by
+      intro R' hR'
+      have hR'pos : 0 < R' := by linarith [dist_nonneg (x := x) (y := x'), hR'.1]
+      have : ∫ (y : X) in Annulus.co x' R₁ R₂, K x' y * f y = (∫ (y : X) in Annulus.oo x' R' R₁, K x' y * f y) +
+          (∫ (y : X) in Annulus.co x' R₁ R₂, K x' y * f y) - ∫ (y : X) in Annulus.oo x' R' R₁, K x' y * f y := by
+        simp
+      rw [diff_eq_compl_inter, inter_comm, ← Annulus.co_eq, this]
+      have : Annulus.oo x' R' R₂ = Annulus.oo x' R' R₁ ∪ Annulus.co x' R₁ R₂ := by
```
grunweg: These are the same lemmas as above, right? So you could extract them into a separate lemma, and simplify both branches accordingly.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -943,24 +943,98 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
...
+    -- apply continuity
+    have le_R1 : ‖∫ (y : X) in Annulus.oo x' R₁ R₂, K x' y * f y‖ₑ ≤
+        ‖∫ (y : X) in Annulus.oo x' R₁ R₁, K x' y * f y‖ₑ + sup := by
+      refine ContinuousWithinAt.closure_le ?_ ?_ ?_ this
+      · simp [closure_Ioo hR₂.ne, hR₂.le]
+      · apply continuousWithinAt_const
+      · apply ContinuousWithinAt.add ?_ continuousWithinAt_const
+        exact small_annulus_right ha hf |>.continuousWithinAt.enorm
+    simpa using le_R1
+  · have : ∀ (R' : ℝ), R' ∈ Ioo (dist x x') R₁ → ‖∫ (y : X) in ball x' R₂ \ ball x' R₁, K x' y * f y‖ₑ ≤
```
grunweg: Did you know about `have`s with parameters: i.e., you can write something like
`have (R' : ℝ) (hR' : R'∈ Ioo (dist x x') R₁) : (rest of your statement)` and drop the `intro R' hR'` line from your proof.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -943,24 +943,98 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
...
+      apply le_trans ?_ <| le_iSup _ (i := R₂)
+      apply le_trans ?_ <| le_iSup _ (i := hR'.2.trans hR₂)
+      apply le_trans ?_ <| le_iSup _ (i := x')
+      apply le_trans (by rfl) <| le_iSup _ (i := hR'.1)
+    -- apply continuity
+    have le_R1 : ‖∫ (y : X) in ball x' R₂ \ ball x' R₁, K x' y * f y‖ₑ ≤
+        ‖∫ (y : X) in Annulus.oo x' R₁ R₁, K x' y * f y‖ₑ + nontangentialOperator K f x := by
+      refine ContinuousWithinAt.closure_le ?_ ?_ ?_ this
+      · simp [closure_Ioo hx'.ne, hx'.le]
+      · apply continuousWithinAt_const
```
grunweg, line 1029: Same comments as above, about continuity.

grunweg (commented):
Thanks! This code seems safe to merge, since this directory does not enforce mathlib standards anyway. In case you're interested, I left some comments indicating areas where this *could* be shortened.
I'll let Floris take a look as well, so he can decide whether to merge it.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
JasperMS: Yes this is deliberate

Carleson/TwoSidedCarleson/NontangentialOperator.lean
JasperMS: Wouldn't it be nice... If you know a way, I want to know as well!
In the L2Estimate file there is a similar approach so maybe this is just how it's done...?

Carleson/TwoSidedCarleson/NontangentialOperator.lean
JasperMS: I'm afraid I don't understand what you mean.
These goals result from `rw` on the line above and without naming them they carry down until the main flow is done (or require swapping), which I find counterintuitive.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
JasperMS: The lemmas are not identical and `this` is used twice, so it's a convenience. The base lemma is `oo_union_co` and is already in `ToMathlib`.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
JasperMS: I know, but will postpone that until 10.1.7 drops so that we can be sure the definitive version continues to slot in how we expect it to.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
JasperMS: No I didn't, thanks!

Carleson/TwoSidedCarleson/NontangentialOperator.lean
JasperMS: I tried and it gives unexpected problems, so I'll leave it like this. Maybe it is because the type is only inferred after `this` is resolved?

Carleson/TwoSidedCarleson/NontangentialOperator.lean
grunweg: Took a look locally: you can use `gcongr with R₁ hR₁ R₂ hR₂ x' hx'` instead.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
grunweg: You're right, talking about `refine` was non-sense.
You can use the `swap` (or `rotate_left`) tactic to change the order of the goals, though. I just pushed a commit inlining the first goal (i.e., proving it as part of the `have`), and using `swap` for the other one. I you prefer it otherwise, I can also revert that one.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
grunweg: I tried, and didn't find an obvious way: let's merge as-is.

Carleson/TwoSidedCarleson/NontangentialOperator.lean
grunweg: ... except that this will change the goal to something ball \ ball = annulus, whereas the statement is only true with ball \ closedBall. For a mathlib proof, I'd try harder to make this shorter. Here, let's just merge this and move onwards to the next lemma.

grunweg:
You're welcome. I'm happy with merging this PR, so let me do this now.

PR #398, opened 2025-06-19
Lemma 10.1.7

JasperMS:
With the correct statement in place in 10.1.8, it was feasible to prove both estimates without resorting to epsilon-delta intricacies.
Instead `continuousWithinAt_of_dominated` was used.

This led to intricacies because continuity of an indicator had to be proved. This was feasible but only at the boundary. Therefore no `ContinuousOn` or `Continuous` statement was shown.
In fact if you look precisely the formalised statement aligns with the blueprint in the sense that they show the same thing.

In terms of the formalisation, I struggled to play nicely with `nhdsWithin` on the reals, so it might be that some statements can be optimised. Also the proofs look similar, but in critical parts the left-or-rightness is being used so I'm afraid it cannot be easily unified.

### DIFF
--- a/Carleson/TwoSidedCarleson/Basic.lean
+++ b/Carleson/TwoSidedCarleson/Basic.lean
@@ -26,6 +26,43 @@ lemma memLp_top_K_on_ball_complement (hr : 0 < r) {x : X}:
       · intro y hy
         apply enorm_K_le_ball_complement' hr hy
 
+lemma czoperator_bound {g : X → ℂ} (hg : BoundedFiniteSupport g) (hr : 0 < r) (x : X) :
+    ∃ (M : ℝ≥0), ∀ᵐ y ∂(volume.restrict (ball x r)ᶜ), ‖K x y * g y‖ ≤ M := by
+  let M0 := (C_K a / volume (ball x r) * eLpNorm g ∞).toNNReal
+  use M0
+  rw [ae_iff, Measure.restrict_apply₀']
+  · let M1 := (C_K a / volume (ball x r)).toNNReal
+    let M2 := (eLpNorm g ∞).toNNReal
+    have : { y | ¬‖K x y * g y‖ ≤ M0} ⊆ { y | ¬‖K x y‖ ≤ M1 ∨ ¬‖g y‖ ≤ M2} := by
+      rw [setOf_subset_setOf]
+      intro y
+      contrapose!
+      intro hy
+      rw [norm_mul]
+      trans M1 * M2
+      · apply mul_le_mul hy.left hy.right
+        case b0 | c0 => simp only [norm_nonneg, NNReal.zero_le_coe]
+
+      apply le_of_eq
+      norm_cast
+      rw [← toNNReal_mul]
+    rw [← Measure.restrict_apply₀']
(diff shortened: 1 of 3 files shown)


Carleson/TwoSidedCarleson/NontangentialOperator.lean
```diff
@@ -942,20 +942,125 @@ theorem simple_nontangential_operator_le (ha : 4 ≤ a)
...
+omit [CompatibleFunctions ℝ X (defaultA a)] [IsCancellative X (defaultτ a)] in
 /-- Part of Lemma 10.1.7, reformulated -/
-theorem small_annulus_left (ha : 4 ≤ a)
-    {f : X → ℂ} (hf : BoundedFiniteSupport f) {R₂ : ℝ} :
-    Continuous (fun R₁ ↦ ∫ y in Annulus.oo x' R₁ R₂, K x' y * f y) := by
-  sorry
+theorem small_annulus_left {g : X → ℂ} (hg : BoundedFiniteSupport g) {R₁ R₂ : ℝ} (hR₁ : 0 ≤ R₁):
+    ContinuousWithinAt (fun R ↦ ∫ y in Annulus.oo x R R₂, K x y * g y) (Ioo R₁ R₂) R₂ := by
+  by_cases hR1R2 : R₁ < R₂
+  case neg => rw [Ioo_eq_empty hR1R2, ContinuousWithinAt, nhdsWithin_empty]; exact Filter.tendsto_bot
```
grunweg, line 1008: FWIW, I'd use a focusing dot (and `swap` if needed): but this need not block merging this PR.

Carleson/TwoSidedCarleson/Basic.lean
```diff
@@ -26,6 +26,43 @@ lemma memLp_top_K_on_ball_complement (hr : 0 < r) {x : X}:
...
+    · apply measure_mono_null_ae this.eventuallyLE
+      rw [setOf_or]
+      apply measure_union_null
+      · rw [← ae_iff]
+        apply ae_restrict_of_forall_mem measurableSet_ball.compl
+        intro y hy
+        simp_rw [M1, ← ENNReal.toReal.eq_1, ← toReal_enorm]
+        apply (ENNReal.toReal_le_toReal enorm_lt_top.ne ?_).mpr
+        · apply enorm_K_le_ball_complement hy
+        · exact (div_lt_top coe_ne_top ((measure_ball_pos volume x hr).ne.symm)).ne
```
grunweg, line 59: Can `finiteness` do this now?

grunweg (commented):
Thanks, looks good. I left two comments, which are optional: let me just merge this (and feel free to clean this up in later PRs).

grunweg:
CI is pretty green, so let's merge this already.

Carleson/TwoSidedCarleson/Basic.lean
JasperMS: FYI It can't, it's stuck on measure_ball_pos which isn't tagged with `positivity`.

grunweg: So you could do the following, right?
```
have := measure_ball_pos volume x hr
finiteness

PR #395, opened 2025-06-17
Proof of Lemma 10.2.5

js2357:
- I had to weaken the conclusion of `tsum_volume_czBall3_le` because `maximal_theorem` (Lemma 10.2.1) is formalized with a weaker constant than appears in the blueprint. I changed the statement and proof of Lemma 10.2.5 in the blueprint, but I didn't change Lemma 10.2.1 because I wasn't prepared to fix its proof.
- I had to strengthen the assumptions in `eLpNorm_czRemainder_le` and `tsum_eLpNorm_czRemainder_le`, but that's because the original formalization was missing an assumption present in the blueprint, so I don't think that should cause any problems.

PR #401, opened 2025-06-19
Proof of Lemma 10.2.6

js2357:
Proof of Lemma 10.2.6 (`estimate_good`). Includes material from the proof of Lemma 10.2.5 (#395).

- I fixed the conclusion of the lemma, which was incorrectly formalized using `czApproximation f ha α` rather than `czApproximation f ha (c10_0_3 a * α)` (corresponding to the fact that `α'` plays the role of `α` in the blueprint).
- The formalization of the lemma didn't include the assumption `hT` from Lemma 10.0.3, so I added it. As far as I know, Lemma 10.2.6 is only used to prove Lemma 10.0.3, so I don't think there's any harm in that.
- In order to apply `hT`, we needed to know that the `czApproximation` of a function of bounded finite support is also of bounded finite support. This was previously not stated in Lemma 10.2.5, so I added the extra statement and proof to the blueprint.

### DIFF
--- a/Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
+++ b/Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
@@ -558,7 +558,8 @@ private lemma MeasurableSet.czPartition (hX : GeneralCase f α) (i : ℕ) :
     MeasurableSet (czPartition hX i) := by
   refine i.strong_induction_on (fun j hj ↦ ?_)
   unfold _root_.czPartition
-  measurability
+  apply measurableSet_ball.diff ∘ (MeasurableSet.biUnion (to_countable _) hj).union
+  exact MeasurableSet.biUnion (to_countable _) (fun _ _ ↦ measurableSet_ball)
 
 lemma czBall_subset_czPartition {hX : GeneralCase f α} {i : ℕ} :
     czBall hX i ⊆ czPartition hX i := by


Carleson/TwoSidedCarleson/Basic.lean
```diff
@@ -73,11 +73,16 @@ lemma czOperator_aestronglyMeasurable {g : X → ℂ} (hg : BoundedFiniteSupport
...
   apply AEStronglyMeasurable.indicator
   · apply Continuous.comp_aestronglyMeasurable₂ (by fun_prop) aestronglyMeasurable_K
-    exact hg.aestronglyMeasurable.comp_snd
+    exact hg.comp_snd
   · conv => arg 1; change {x : (X × X) | x.2 ∈ (ball x.1 r)ᶜ}
     simp_rw [mem_compl_iff, mem_ball, not_lt]
     apply measurableSet_le <;> fun_prop
 
+@[fun_prop]
+lemma czOperator_aestronglyMeasurable {g : X → ℂ} (hg : BoundedFiniteSupport g) :
```
grunweg, line 83: Does `g` being `AEStronglyMeasurable` also suffice here?

Carleson/ToMathlib/WeakType.lean
```diff
@@ -166,6 +166,14 @@ lemma continuousWithinAt_distribution (t₀ : ℝ≥0∞) :
                 ≤ distribution f t₀ μ := distribution_mono_right (le_of_lt h.2)
               _ ≤ distribution f t₀ μ + ε := le_self_add
 
+lemma distribution_pow (ε : Type*) [SeminormedRing ε] [NormOneClass ε] [NormMulClass ε] (f : α → ε)
```
grunweg, line 171: These classes only apply to norms, not enorms, right? (If changing this involves adding new enorm versions of these classes to mathlib, then feel free to leave it to e.g. me.)

Carleson/TwoSidedCarleson/Basic.lean
js2357: Yes, that's the lemma immediately before the one you commented on (which I just added). I needed the version that only assumes `AEStronglyMeasurable`, but the pre-existing version was the version that assumed `BoundedFiniteSupport`. I kept the `BoundedFiniteSupport` version around for now to avoid breaking stuff, but I think it makes sense to remove it at some point.

Carleson/ToMathlib/WeakType.lean
js2357: I'm definitely not an expert on the technicalities here, but I think you're right that trying to make this any more general would involve adding new enorm versions of the classes.

grunweg: Fair point. Can you add a comment about this (such as the following)? (I've rewritten much of the WeakType file to only work with enorms, hence having a comment on this recent addition would be helpful to me.)
```suggestion
-- TODO: investigate generalising this to enorm classes
-- This requires adding enorm versions of NormOneClass and NormMulClass
lemma distribution_pow (ε : Type*) [SeminormedRing ε] [NormOneClass ε] [NormMulClass ε] (f : α → ε)

File: Carleson/TwoSidedCarleson/Basic.lean (hunk 1 above)
Comment (grunweg) [reply Carleson/TwoSidedCarleson/Basic.lean:83]: Sure, this can always be done later. How about adding a comment, like this?
```suggestion
-- TODO: remove in favour of `czOperator_aestronglyMeasurable` (and unprime that one)
@[fun_prop]

**Comment by js2357** [comment]
I think this should be working now.

One last thing: every instance of `[CompatibleFunctions ℝ X (defaultA a)]` in WeakCalderonZygmund.lean was being used solely to imply `[Nonempty X]`, so I changed them all to `[Nonempty X]`. If there was some technical reason for using `CompatibleFunctions`, let me know and I'll change them back.

File: Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean [hunk 3]
```diff
@@ -557,8 +558,7 @@ private lemma MeasurableSet.czPartition (hX : GeneralCase f α) (i : ℕ) :
     MeasurableSet (czPartition hX i) := by
   refine i.strong_induction_on (fun j hj ↦ ?_)
   unfold _root_.czPartition
-  apply measurableSet_ball.diff ∘ (MeasurableSet.biUnion (to_countable _) hj).union
-  exact MeasurableSet.biUnion (to_countable _) (fun _ _ ↦ measurableSet_ball)
+  measurability
```
grunweg: This reverts an intentional change from #395 --- I guess a bad merge. Let me fix that quickly.
```suggestion
  apply measurableSet_ball.diff ∘ (MeasurableSet.biUnion (to_countable _) hj).union
  exact MeasurableSet.biUnion (to_countable _) (fun _ _ ↦ measurableSet_ball)

PR #409, opened 2025-06-24
Fix several sorrys in `RealInterpolation`

Parcly-Taxel:


Carleson/ToMathlib/RealInterpolation/Misc.lean
```diff
@@ -642,10 +644,12 @@ lemma truncCompl_anti {x : α} (hab : t ≤ s) (hf : ‖trunc f t x‖ₑ ≠ 
   exact foo hf (trunc_mono hab) obs
 
 /-- The norm of the complement of the truncation is antitone in the truncation parameter -/
-lemma eLpNorm_truncCompl_anti (hf : eLpNorm f 1 μ ≠ ⊤) :
+lemma eLpNorm_truncCompl_anti (hf : eLpNorm f 1 μ ≠ ⊤) (mf : AEStronglyMeasurable f μ) :
```
grunweg, line 648: Note: I'm not sure if this is the correct fix --- perhaps fixing downstream proofs will tell.
Can you add a comment about this, please (so whoever encounters is aware)? Thanks!
```suggestion
-- XXX: the conditions `hf` and `mf` may need to be tweaked
lemma eLpNorm_truncCompl_anti (hf : eLpNorm f 1 μ ≠ ⊤) (mf : AEStronglyMeasurable f μ) :

grunweg: (To some extent, this is pre-existing.)

fpvandoorn:
Thanks!

PR #410, opened 2025-06-24
chore: golf lebesgue_differentiation

js2357:
Construct the sequence `ρ` increasing to `r` using `exists_seq_strictMono_tendsto r` rather than an explicit formula.


grunweg:
Thanks, that's nicer indeed!

PR #411, opened 2025-06-24
fix interpolated_exponents

mo271:
There was a mistake in `switch_exponents`, which lead to it being impossible to prove and more failures below.


mo271:
ah, I see #409 already fixed that...

PR #413, opened 2025-06-25
Lemma 3.0.3

Parcly-Taxel:
All the `<` under the suprema were changed to `≤` – and the supremum structure was changed to two `biSup`s rather than five `iSup`s – but this should all be without loss of generality.


fpvandoorn:
Thanks!

PR #417, opened 2025-06-26
Lemma 10.0.2

JasperMS:
The result has been shown up to a variant of `lintegral_iSup'`, namely `eLpNorm_iSup'`.
The exact statement and formalization of that result is something I have put off for now, but at least it should hold for `p = 2` which is how it is used here.

It includes the formalization of measurability, analogous to `simpleNontangentialOperator`.


fpvandoorn:
Oh, that is a good lemma to add to `ToMathlib`, and probably requires some other new results (e.g. swapping suprema and ENNReal-powers). I'll keep task 127 open to mean "solve this sorry".

In the next PR, please indeed also fix the statement/proof of Lemma 10.0.2 to account for the changing constant.

PR #421, opened 2025-06-28
Lemma 10.2.8, 10.2.9 and 10.0.3

Parcly-Taxel:


js2357:
I don't think this is quite right. The blueprint says that α' = cα plays the role of α in the discussion after Lemma 10.2.5. E.g., every `(hX : GeneralCase f α)` should actually have been `(hX : GeneralCase f (c10_0_3 a * α))`, and similarly for every `czRemainder f α`, etc. This presumably explains why you ended up with a different constant `C10_2_9`.

I defined a function `α'` while I was proving Lemma 10.2.7; you can what I did [here](https://github.com/js2357/carleson/tree/Lemma10.2.7).

Parcly-Taxel:
> The blueprint says that α' = cα plays the role of α in the discussion after Lemma 10.2.5.

No, that replacement only applies to the `α`s used to construct the good and bad functions (`czApproximation`, `czRemainder`). So `estimate_good`, which you formalised yourself, is correct to replace `α` with `α'` in `czApproximation`:
```lean
lemma estimate_good [Nonempty X] (hf : BoundedFiniteSupport f)
    (hα : ⨍⁻ x, ‖f x‖ₑ / c10_0_3 a < α)
    (hT : HasBoundedStrongType (czOperator K r) 2 2 volume volume (C_Ts a)) :
    distribution (czOperator K r (czApproximation f (c10_0_3 a * α))) (α / 2) volume ≤
    C10_2_6 a / α * eLpNorm f 1 volume := by
```
but `GeneralCase f α` is unaffected.

I am still waiting for a proof of `aemeasurable_V` in this PR.

Parcly-Taxel:
> Also, I think there are at least two other mistakes in the formalization, due to a separate notation change. The balls B_j after Lemma 10.2.5 are the balls that were originally called B_j*, i.e., B_j should correspond to `czBall3`.
> 
>     * In the definition of `czOperatorBound`: the `(czRadius hX i).toNNReal` should be `(3 * czRadius hX i).toNNReal`. (The easiest way to see this is to note that r_i is the radius of B_i, and B_i is correctly formalized as `czBall3 hX i`.)
> 
>     * Every `czBall2` should actually be `czBall6` (with the obvious change in definition).

This is a curious mistake in the blueprint, not the Lean code. `B_j` is redefined as `B_j*`, yes, but then it is immediately redefined as the original definition `ball (czCenter hX i) (czRadius hX i)`. The redefinition of `B_j` as `B_j*` should be deleted.

The blueprint reflects the code, not the other way around.

js2357:
> This is a curious mistake in the blueprint, not the Lean code. `B_j` is redefined as `B_j*`, yes, but then it is immediately redefined as the original definition `ball (czCenter hX i) (czRadius hX i)`. The redefinition of `B_j` as `B_j*` should be deleted.

Are you interpreting (10.2.38) as a redefinition of the notation B_j? I don't think that's what it is; it's a new definition of x_j and r_j, based on the new definition of B_j.

Your interpretation doesn't work with the proof. For example, the proof of Lemma 10.2.7 uses the fact that supp b_j ⊆ B_j, which is not necessarily true of the original B_j, but is true of B_j*.

Also, you ignored my comment about the definition of `czOperatorBound`, which formalized r_j as `czRadius hX i` and B_j as `czBall3 hX i`. I think we both agree that r_j is the radius of B_j, whatever B_j is, so the definition is certainly inconsistent with the blueprint one way or the other.

> Here you're not reading the blueprint closely enough. Before the phrase "with `α'` replacing `α`", `α` in its original meaning is used in a special case of the Lemma 10.0.3 proof.

I'm referring to your proof of Lemma 10.2.8, which is not part of that special case.

Parcly-Taxel:
> Considering that we have four different radii to keep track of for each j, it might be simplest to move away from the stars entirely and do something like B_{1,j}, B_{3,j}, B_{6,j}, B_{7,j}.

I'm happy with this. I'd like it if you also update Lemma 10.2.7's proof in the blueprint to use these new names (then I will see to the rest of the section).

fpvandoorn:
> Sounds good. Even before this conversation, I was already thinking that the blueprint could be written much more clearly. Considering that we have four different radii to keep track of for each j, it might be simplest to move away from the stars entirely and do something like B_{1,j}, B_{3,j}, B_{6,j}, B_{7,j}.

Feel free to do this! I also found the current formulation not that clear, so please go ahead.

PR #422, opened 2025-06-28
feat: integrableOn_K_mul

js2357:
Prove `integrableOn_K_mul`, which will be needed for the proof of Lemma 10.2.7.

Use it to simplify the proofs of `integrable_Ks_x` and `integrableOn_K_Icc`.


fpvandoorn:
Thanks!

PR #423, opened 2025-07-01
Lemma 7.2.2 (correctly)

Parcly-Taxel:
I still have one problem that I cannot solve: how is
```
volume (EAnnulus.cc x' (ENNReal.ofReal R₂) (ENNReal.ofReal R₂)) = 0 ?
```
Do I need any further assumptions?


fpvandoorn:
Thanks a lot, looks good!
In the blueprint, in the last sentence of the proof, can you also fix the reference `\eqref{nontanbound}` to the right inequality?

PR #425, opened 2025-07-01
Proof of Lemma 10.2.7

js2357:
One TODO left in a comment to figure out where a new ToMathlib lemma should go, but otherwise this is finished. I don't think the lemma has an obvious home, so suggestions are welcome.

I made the blueprint changes to get rid of the B_j* / B_j change in notation (and fix a couple unrelated typos).


Carleson/ToMathlib/Data/Set/Card.lean
```diff
@@ -0,0 +1,12 @@
+import Mathlib.Data.Set.Card
+
+namespace Set
+
+variable {α : Type*}
+
+lemma encard_subtype_le (P Q : α → Prop) :
+    {x : Subtype P | Q ↑x}.encard ≤ {x | Q x}.encard :=
```
fpvandoorn, line 8: The more natural statement for a lemma with this name is 
`{x | Q x}.encard ≤ α.encard`, c.f. [Finite.card_subtype_le](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Finite/Card.html#Finite.card_subtype_le). Could you rename this, maybe to `encard_subtype_le_encard_subtype`?
Note: this might also be more naturally stated using [Set.encard](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Set/Card.html#Set.encard), because then this is `(A ↓∩ B).encard ≤ B.encard`.

fpvandoorn (commented):
Thanks, looks good. I have one comments, but I'll merge it now as-is. If you agree with the comment, please incorporate it in a follow-up PR.

Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
```diff
@@ -1051,27 +1111,439 @@ lemma estimate_good [Nonempty X] {hf : BoundedFiniteSupport f}
...
+  refine iUnion_subset (fun j y hy ↦ ?_)
+  rw [mem_ball'] at hy
+  have : 6 * czRadius hX j ≤ dist x (czCenter hX j) := six_mul_czRadius_le_of_mem_Ω hx hX j
+  have hj := Subtype.coe_prop j
+  simp only [𝒥₂, tsub_le_iff_right, mem_setOf_eq] at hj
+  constructor
+  · linarith [dist_triangle_right x (czCenter hX j) y]
+  · linarith [dist_triangle x (czCenter hX j) y]
+
+-- TODO: Move to appropriate ToMathlib file
```
fpvandoorn, line 1360: Oh, it's probably good to put this in `ToMathlib`. Does `#find_home` suggest something reasonable? Do you need something that is not imported in `Mathlib.MeasureTheory.Measure.Restrict`? If so, a new file might be ok.

js2357: `#find_home` and `find_home!` only find the current file. `Set.encard` isn't available in `Mathlib.MeasureTheory.Measure.Restrict`. I guess it makes sense that there isn't any file considering the interaction of measures and `encard`, or else something like this lemma would probably already exist.

Any idea what to call the new file?

Carleson/ToMathlib/Data/Set/Card.lean
js2357: The name change makes sense. I'm not sure I follow the note. The `encard` in the lemma is already `Set.encard`; do you want me to change the statement to use the `↓∩` notation? If we do that, should we pick a different name? It seems like lemmas about `↓∩`  use `preimage_val` in their names. I guess we should also make `left` and `right` versions while we're at it. Maybe something like this?



```lean
import Mathlib.Data.Set.Card

namespace Set

variable {α : Type*}

open Notation in
lemma encard_preimage_val_le_encard_left (P Q : Set α) : (P ↓∩ Q).encard ≤ P.encard :=
  (Function.Embedding.subtype _).encard_le 

open Notation in
lemma encard_preimage_val_le_encard_right (P Q : Set α) : (P ↓∩ Q).encard ≤ Q.encard :=
  Function.Embedding.encard_le ⟨fun ⟨⟨x, _⟩, hx⟩ ↦ ⟨x, hx⟩, fun _ _ h ↦ by
    simpa [Subtype.coe_inj] using h⟩

end Set
```

Carleson/TwoSidedCarleson/WeakCalderonZygmund.lean
fpvandoorn: I don't think it's unreasonable to add imports so that we can talk about `Set.encard` in measure theory (it's used in the file about the counting measure). But it might be easier to just add a new file `SumRestrict` or something.

PR #427, opened 2025-07-02
Proof of revised Lemma 6.3.4

mariainesdff:


Carleson/Defs.lean
```diff
@@ -626,7 +626,7 @@ variable [ProofData a q K σ₁ σ₂ F G]
 include a q K σ₁ σ₂ F G
 
 variable (X) in
-lemma S_spec : ∃ n : ℕ, (∀ x, -n ≤ σ₁ x ∧ σ₂ x ≤ n) ∧
+/- lemma S_spec : ∃ n : ℕ, (∀ x, -n ≤ σ₁ x ∧ σ₂ x ≤ n) ∧
```
Parcly-Taxel: Please do not change `Defs.lean` in this PR. In #433 I tweak the definition and blueprint so that `S` must be positive, and then you can use `S_pos : 0 < S`.

mariainesdff: I changed `S_spec` here after discussing this PR with @fpvandoorn. It seems from your commit message that you knew that I had proved positivity of `S`, so I do not understand why you reproved it.

Parcly-Taxel: There is nothing to prove here about choosing `S` to be positive – it never hurts to increase `S`.

I reproved it because I want the code to be neat, and your current modifications in `Defs.lean` are quite sloppy.

grunweg: :butterfly: 
@Parcly-Taxel Please be more mindful in how you choose your words. There is no need to be personal ("your code").

fpvandoorn: Jeremy, I appreciate the work you do for the Carleson project, but you need to be respectful to other contributors as well. You shouldn't call other people's code sloppy. It's also not very nice to ask people to undo their work, because you also did it.

If you think you can give a nicer proof of something, either leave a review comment with a positive message, or wait until the PR is merged and make a PR changing the proof afterwards.

grunweg: One additional consideration is that most people working on this project are not doing so full-time.
Picking a task others have claimed and completing it faster (without coordinating) is not nice.
(In fact, this can be rather demotivating --- and I expect it to slow down progress on Carleson overall.)

Parcly-Taxel: Right then. I apologise for calling the code here sloppy.

What I was getting at was that since increasing `S` preserves all existing properties, it is easier when adding the `0 < S` condition to just take an `S` satisfying the existing conditions and add 1, rather than trying to prove everything at once.

fpvandoorn: Thanks for your apology.

I'm glad that you want to improve people's proofs. You can do that by writing a suggestion in a review comment, or (if the change is too big to make that comfortable), you can even open a PR to be merged into this PR.

PR #433, opened 2025-07-07
Mark Hardy–Littlewood as finished

Parcly-Taxel:
After filling in a sorry and tidying up a bit I have checked using `#print axioms` that each of the five Lean statements corresponding to Proposition 2.0.6 – as well as `row_correlation` – depends only on `[propext, Classical.choice, Quot.sound]`.


fpvandoorn:
First commit looks good to me. Second commit conflicts unnecessarily with #427.

Parcly-Taxel:
@fpvandoorn OK, I have reverted the second commit (but have saved it locally so I can apply it later, after #427 is done).

fpvandoorn:
Thanks!

PR #445, opened 2025-07-10
feat: proof of Lemma 11.3.5

sgouezel:
I had to change a little bit the blueprint and the Lean statements, as the original proof was only correct up to sign: in the proof of Lemma 11.3.5, one wants to approximate a function kappa of the form .../(1 - e^{i x}), but the proof pretends that it's .../(1 - e^{-ix}) instead, see (11.3.24). In the PR, I change the definition of the approximate Hilbert transform (11.3.2), replacing M_{-n-N} S_{n+N} M_{n+N} with M_{n+N} S_{n+N} M_{-n-N} to make sure that what comes out is indeed .../(1 - e^{ix}). With this modified approximate Hilbert transform, everything seems to work fine.


fpvandoorn:
Thanks, this looks great! Thanks for spotting and fixing the mistake in the blueprint.

PR #452, opened 2025-07-13
feat: Lemma 11.1.6

sgouezel:
I had to modify a little bit the blueprint, as it claimed an inequality was always true, but I could only prove it in some range. If one writes down the details of the proof, only this range is needed, so the global strategy of proof works perfectly well modulo these minor adjustments.

I made the following Lean change: the `L^2` norm of `f` on an interval `[a, b]` was used as `eLpNorm (indicator (Icc a b) f) 2` (where there is a measure there, the volume, which is implicit). I changed it to `eLpNorm f 2 (volume.restrict (Icc a b))` here and there, as this is how we do integrals on sets and it is much more convenient to use since we have a big API for restrictions of measures (and avoiding indicators is always a good idea as they are pretty painful).

### DIFF
--- a/Carleson/ToMathlib/MeasureTheory/Integral/Periodic.lean
+++ b/Carleson/ToMathlib/MeasureTheory/Integral/Periodic.lean
@@ -106,8 +106,9 @@ theorem map_coe_addCircle_volume_eq :
   ext s hs
   simp [hs]
 
-theorem map_coe_addCircle_volume_ac :
-    (Measure.map (fun (x : ℝ) ↦ (x : AddCircle T)) volume) ≪ volume := by
+theorem quasiMeasurePreserving_coe_addCircle :
+    QuasiMeasurePreserving (fun (x : ℝ) ↦ (x : AddCircle T)) := by
+  refine ⟨by fun_prop, ?_⟩
   rw [map_coe_addCircle_volume_eq]
   exact smul_absolutelyContinuous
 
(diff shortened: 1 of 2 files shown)


Carleson/ToMathlib/MeasureTheory/Integral/Periodic.lean
```diff
@@ -72,24 +72,45 @@ open AddCircle
...
+  rw [this, Measure.map_sum (by fun_prop)]
+  have A (n : ℤ) : Measure.map (fun (x : ℝ) ↦ (x : AddCircle T))
+      (volume.restrict (Ioc (n • T) ((n + 1) • T)) ) = volume := by
+    simp only [zsmul_eq_mul, Int.cast_add, Int.cast_one, add_mul, one_mul]
+    exact (AddCircle.measurePreserving_mk T (n * T)).map_eq
+  simp only [A]
+  ext s hs
+  simp [hs]
+
+theorem map_coe_addCircle_volume_ac :
```
fpvandoorn: formulate in terms of [MeasureTheory.Measure.QuasiMeasurePreserving](https://leanprover-community.github.io/mathlib4_docs/Mathlib/MeasureTheory/Measure/QuasiMeasurePreserving.html#MeasureTheory.Measure.QuasiMeasurePreserving)?

fpvandoorn (commented):
Great, thanks a lot! One comment (I'm also happy to merge it as-is).

Carleson/ToMathlib/MeasureTheory/Integral/Periodic.lean
sgouezel: Good idea, done.

PR #457, opened 2025-07-15
remove regularity assumption

fpvandoorn:
Sigma-finite is also redundant.


Carleson/ToMathlib/DoublingMeasure.lean
```diff
@@ -476,7 +476,6 @@ instance : IsOpenPosMeasure (volume : Measure X) := isOpenPosMeasure_of_isDoubli
 -- #synth SecondCountableTopology X
 -- #synth SeparableSpace X
```
Parcly-Taxel: You should remove all these `#synth`s as well in this PR.

fpvandoorn: Good idea

PR #458, opened 2025-07-15
feat: optimize the constants

sgouezel:
Instead of lettind `D = 2 ^ {100 * a^2}`, let `D = 2 ^ {c * a^2}` for some fixed natural `c`, and express all the constants in terms of `c` (still keeping all the relevant constants of the form `2 ^ {n * a^2}` for `n = n(c)`, because it wouldn't make sense to overoptimize). For `c = 100`, we should recover essentially the constants in the blueprint. However, we can take `c = 7`, which gives better constants in the end.

NB: this depends on branches completing the proof, so this should not be merged for now: I'll merge master and clean up my branch once everything is settled.


fpvandoorn:
Nice!
I'm not completely sure about merging this into master at all. It might be nice to have the blueprint constants aligned with the formalization constants. I'm happy to put this in e.g. a branch we link to from the README.

sgouezel:
In fact there's something funny that I realized when doing this: c shouldn't be too small (it can go down up to 7), but it shouldn't be too large either (I think more than 200 breaks down at some other place). This argument is really subtle!

It's a good idea to have just two lemmas about `c`, and never unfold it later on. I'll do it like that.

docs/_includes/mathjax.html
```diff
@@ -6,5 +6,5 @@
   });
 </script>
 <script type="text/javascript" async
-  src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
+  src="https://𝕔.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
```
skaslev: Unintentional change?

sgouezel: Yes, definitely! Search and replace epic fail :-)

Carleson/Defs.lean
```diff
@@ -306,7 +306,20 @@ lemma carlesonOperator_const_smul [FunctionDistances ℝ X] (K : X → X → ℂ
...
+in the end. We will formalize it using a parameter `𝕔` (that we fix equal to `100` to follow
+the blueprint) and having `D = 2 ^ (𝕔 * a ^ 2)`. We register two lemmas `seven_le_c` and
+`c_le_111` and will never unfold `𝕔` from this point on. -/
+--irreducible_def 𝕔 : ℕ := 7
+
+--lemma seven_le_c : 7 ≤ 𝕔 := by simp [𝕔]
+--lemma c_le_111 : 𝕔 ≤ 111 := by simp [𝕔]
+
+lemma exists_c : ∃ (c : ℕ), 7 ≤ c ∧ c ≤ 111 := ⟨7, le_rfl, by norm_num⟩
+def 𝕔 : ℕ := exists_c.choose
```
fpvandoorn, line 323: This is a nice solution.

Carleson/Discrete/ForestComplement.lean
```diff
@@ -1063,7 +1063,7 @@ def C5_1_3_optimized (a : ℕ) (q : ℝ≥0) := C2_0_3 a q * 2 ^ (29 * a + 23) /
 
 /-- The constant used in Lemma 5.1.3 in the blueprint,
 with value `2 ^ (131 * a ^ 3) / (q - 1) ^ 5` -/
-def C5_1_3 (a : ℕ) (q : ℝ≥0) : ℝ≥0 := 2 ^ (131 * a ^ 3) / (q - 1) ^ 5
+def C5_1_3 (a : ℕ) (q : ℝ≥0) : ℝ≥0 := 2 ^ ((𝕔 + 11 + 𝕔 / 8) * a ^ 3) / (q - 1) ^ 5
```
grunweg, line 1066: This slightly optimises a constant (but that seems fine).

sgouezel: Yeah, some constants were obviously suboptimal. I optimized those to the best value given by the current proof, without changing the proof -- and with the rule that all constants should be of the form `2^{n a^3}` for an integer `n`.

Carleson/Classical/ClassicalCarleson.lean
```diff
@@ -211,3 +211,5 @@ theorem classical_carleson {f : ℝ → ℂ} (cont_f : Continuous f) (periodic_f
 
 
 end
+
+#print axioms classical_carleson
```
grunweg: Would you like to remove this? My understanding is that the dependent PRs have basically landed now.

Carleson/Discrete/MainTheorem.lean
```diff
@@ -11,10 +11,31 @@ variable {X : Type*} {a : ℕ} {q : ℝ} {K : X → X → ℂ} {σ₁ σ₂ : X
 
 /-- The constant used in Proposition 2.0.2,
 which has value `2 ^ (471 * a ^ 3 + 1) / (q - 1) ^ 5` in the blueprint. -/
-noncomputable def C2_0_2 (a : ℕ) (q : ℝ≥0) : ℝ≥0 := C5_1_2 a q + C5_1_3 a q
+noncomputable def C2_0_2 (a : ℕ) (q : ℝ≥0) : ℝ≥0 :=
+    2 ^ ((3 * 𝕔 + 18 + 5 * (𝕔 / 4)) * a ^ 3) / (q - 1) ^ 5
```
grunweg, line 15: So this constant also gets optimised, right? Nice.

Carleson/GridStructure.lean
```diff
@@ -374,24 +374,28 @@ lemma maxCubes_pairwiseDisjoint {s : Finset (Grid X)} :
 end GridManipulation
 
 /-- The constant appearing in Lemma 2.1.2, `2 ^ {-95a}`. -/
-def _root_.C2_1_2 (a : ℕ) : ℝ := 2 ^ (-95 * a : ℝ)
+def _root_.C2_1_2 (a : ℕ) : ℝ := 2 ^ ((-𝕔 + 5) * a : ℝ)
```
grunweg, line 377: To confirm my understanding: you're weakening this bound to allow smaller values of c --- because most proofs are fine with the current, weaker bound? Makes sense!

grunweg (commented):
Just three more files remaining. Looks great so far, just a few more clarifying questions.

It's great to see the proofs clarified in their argument. While some of them get a little longer, that seems like a fair price for more careful handling of the constants.

grunweg:
(My understanding is that all dependents for this PR have landed, i.e. this could be polished and then merged.)

Carleson/GridStructure.lean
sgouezel: That's the bound that comes out of the proof when you change c. For c = 100, as in the blueprint, you get 95, but for other values of c you get another value. It turns out that this value is still large enough for the proof when c \ge 7.

Carleson/Discrete/MainTheorem.lean
sgouezel: Yes, it gets slightly better.

Carleson/Antichain/TileCorrelation.lean
```diff
@@ -200,7 +203,7 @@ lemma range_support {p : 𝔓 X} {g : X → ℂ} {y : X} (hpy : adjointCarleson
       gcongr -- uses hpos, hle.
 
 /-- The constant from lemma 6.2.3. -/
-def C6_2_3 (a : ℕ) : ℝ≥0 := 2^(8 * a)
+def C6_2_3 (a : ℕ) : ℝ≥0 := 2^ (8 * a)
```
grunweg:
```suggestion
def C6_2_3 (a : ℕ) : ℝ≥0 := 2 ^ (8 * a)

File: Carleson/ForestOperator/Forests.lean [hunk 8]
```diff
@@ -38,39 +38,53 @@ lemma estimate_C7_4_5 {a : ℕ} (n : ℕ) (ha : 4 ≤ a) :
...
+    | base => ring_nf; linarith [seven_le_c]
     | succ k lk ih =>
       rw [mul_add_one, mul_add, mul_add_one, pow_add, show 2 ^ 2 = 3 + 1 by norm_num, mul_add_one,
-        add_mul, add_comm]
+        add_mul, add_comm, add_mul]
       gcongr ?_ + ?_
-      calc
-        _ ≤ 2 ^ (2 * 4) * 3 * 25 := by norm_num
-        _ ≤ _ := by gcongr; exact one_le_two
+      · calc
```
grunweg, line 69: This estimate makes me wonder if the constant in this lemma can be optimised further. (But that may be out of scope of this PR.)

Carleson/ForestOperator/Forests.lean
```diff
@@ -353,74 +365,52 @@ lemma adjointCarlesonRowSum_adjoint
...
+    have : (4 : ℝ) ≤ a := mod_cast a4
+    suffices 1/2 + 2 * (a : ℝ) ≤ a ^ 3 by linarith
+    calc
+      _ ≤ 2 * (a : ℝ) + 2 * a := by linarith
+      _ = 1 * 4 * a := by ring
+      _ ≤ a * a * a := by gcongr; linarith
+      _ = _ := by ring
 
-/--
-The approximation being used here should also depend on the constant used to bound the
```
grunweg, line 401: Do you think these comments are worth keeping in some form?

Carleson/ForestOperator/LargeSeparation.lean
```diff
@@ -934,7 +947,7 @@ lemma limited_scale_impact_second_estimate (hp : p ∈ t u₂ \ 𝔖₀ t u₁ u
         rw [dist_comm] at lt_3 lt_4
         exact calculation_4 (lt_1 := lt_1) (lt_2 := lt_2) (lt_3 := lt_3) (lt_4 := lt_4)
           (three := three) (plusOne := plusOne) (X := X)
-    _ ≤ 2 ^ ((-94 : ℝ) * a) * dist_{𝓘 p} (𝒬 u₁) (𝒬 u₂) := by
+    _ ≤ 2 ^ ((-(𝕔-6) : ℝ) * a) * dist_{𝓘 p} (𝒬 u₁) (𝒬 u₂) := by
```
grunweg:
```suggestion
    _ ≤ 2 ^ ((-(𝕔 - 6) : ℝ) * a) * dist_{𝓘 p} (𝒬 u₁) (𝒬 u₂) := by

File: Carleson/ForestOperator/LargeSeparation.lean [hunk 11]
```diff
@@ -946,7 +959,7 @@ lemma limited_scale_impact_second_estimate (hp : p ∈ t u₂ \ 𝔖₀ t u₁ u
       _ ≤ (2 ^ (a : ℝ)) ^ (6 : ℝ) * dist_{𝔠 p, (D ^ 𝔰 p / 4)} (𝒬 u₁) (𝒬 u₂) :=
         mod_cast cdist_le_iterate (f := (𝒬 u₁)) (g := (𝒬 u₂)) (r := (D ^ (𝔰 p)) / 4)
           (k := 6) (x := 𝔠 p) bigger
-    _ ≤ 2 ^ ((-94 : ℝ) * a) * 2 ^ ((Z : ℝ) * n / 2) := by
+    _ ≤ 2 ^ ((-(𝕔-6) : ℝ) * a) * 2 ^ ((Z : ℝ) * n / 2) := by
```
grunweg:
```suggestion
    _ ≤ 2 ^ ((-(𝕔 - 6) : ℝ) * a) * 2 ^ ((Z : ℝ) * n / 2) := by

File: Carleson/ForestOperator/LargeSeparation.lean [hunk 12]
```diff
@@ -1788,35 +1814,38 @@ lemma holder_correlation_tree (hu₁ : u₁ ∈ t) (hu₂ : u₂ ∈ t) (hu : u
 /-! ### Subsection 7.5.3 and Lemma 7.4.5 -/
 
 /-- The constant used in `lower_oscillation_bound`.
-Has value `2 ^ (Z * n / 2 - 201 * a ^ 3)` in the blueprint. -/
+Has value `2 ^ (Z * n / 2 - (2 * 𝕔 + 1) * a ^ 3)` in the blueprint. -/
```
grunweg: That's overzealous search and replace, right?

Carleson/ForestOperator/Forests.lean
sgouezel: I'm planning a next PR where I will clean up all the values of the constants between the blueprint and the formalization, and adjust all these comments accordingly.

Carleson/ForestOperator/LargeSeparation.lean
sgouezel: Yes. Fixed.

Carleson/ForestOperator/Forests.lean
sgouezel: There are many many places in the argument that could be optimized further (including this one), but I don't think it really makes sense to try to optimize them seriously because in the end the constants will be ridiculous in any case. The idea of the current PR is just to keep the current proofs and see what comes out of it, not to be clever :-)

sgouezel:
Thanks for the review! I think I have taken all your comments into account.

Carleson/ForestOperator/Forests.lean
grunweg: Sure. (My worry is more about you forgetting about this comment, as you deleted it. If this is taken care of in a future PR, I'm happy.)

Carleson/ForestOperator/Forests.lean
grunweg: We may have different thresholds of "being clever", but I agree with your general point :-)

grunweg:
Thanks a lot!

PR #463, opened 2025-07-17
Add missing file for the upstreaming dashboard

Ruben-VandeVelde:


YaelDillies (approved):

grunweg:
Thanks! I can imagine the second category being close to empty each time - so omitting it would also be fine with me. What do you think? Otherwise, this looks good.

PR #464, opened 2025-07-18
`enorm`ify most of the lemmas

Parcly-Taxel:
Remove uses of `‖...‖₊` in favour of `‖...‖ₑ` except for `TileCorrelation.lean` and one instance in `PointwiseEstimate.lean`.

----

I won't have time to continue on this immediately because I'll be on vacation climbing Mount Fuji from 19 to 26 July.

### DIFF
--- a/Carleson/Antichain/TileCorrelation.lean
+++ b/Carleson/Antichain/TileCorrelation.lean
@@ -79,10 +79,9 @@ lemma aux_6_2_3 (s₁ s₂ : ℤ) (x₁ x₂ y y' : X) :
   C2_1_3 a / volume (ball x₂ (D ^ s₂)) *
   (D2_1_3 a / volume (ball x₁ (D ^ s₁)) * (edist y y' ^ τ / (D ^ s₁) ^ τ)) := by
   apply mul_le_mul enorm_Ks_le _ (zero_le _) (zero_le _)
-  convert nnnorm_Ks_sub_Ks_le
+  convert enorm_Ks_sub_Ks_le
   rw [← ENNReal.div_rpow_of_nonneg _ _ (τ_nonneg X)]
-  simp only [defaultτ]; congr
-  simp only [coe_nnreal_ennreal_nndist]
+  simp only [defaultτ]
 
 /-- Equation (6.2.5) of Lemma 6.2.1. -/
 lemma e625 {s₁ s₂ : ℤ} {x₁ x₂ y y' : X} (hy' : y ≠ y') (hs : s₁ ≤ s₂) :
--- a/Carleson/Psi.lean
+++ b/Carleson/Psi.lean
@@ -398,19 +398,6 @@ def C2_1_3 (a : ℕ) : ℝ≥0 := 2 ^ (102 * a ^ 3)
 /-- The constant appearing in part 3 of Lemma 2.1.3. -/
 def D2_1_3 (a : ℕ) : ℝ≥0 := 2 ^ (150 * a ^ 3)
 
-/-- preferably use `kernel_bound` instead. -/
-lemma kernel_bound_old {s : ℤ} {x y : X} :
-    ‖Ks s x y‖₊ ≤ 2 ^ a ^ 3 / (volume (ball x (dist x y))).toNNReal := by
-  change ‖K x y * ψ (D ^ (-s) * dist x y)‖ ≤ 2 ^ a ^ 3 / volume.real (ball x (dist x y))
-  apply le_trans <| calc
-    ‖K x y * ψ (D ^ (-s) * dist x y)‖
-      = ‖K x y‖ * ‖(ψ (D ^ (-s) * dist x y) : ℂ)‖ := norm_mul ..
-    _ ≤ ‖K x y‖ * 1               := by gcongr; rw [norm_real]; exact abs_ψ_le_one D _
-    _ ≤ ‖K x y‖                   := by rw [mul_one]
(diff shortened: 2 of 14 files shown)


Carleson/ToMathlib/HardyLittlewood.lean
```diff
@@ -834,8 +834,7 @@ theorem hasWeakType_maximalFunction_equal_exponents₀ [BorelSpace X]
 theorem hasWeakType_maximalFunction_equal_exponents
     [BorelSpace X] {p : ℝ≥0} (h𝓑 : 𝓑.Countable) (hp : 0 < p) :
     HasWeakType (fun (u : X → E) (x : X) ↦ maximalFunction μ 𝓑 c r p u x)
-      p p μ μ (A ^ ((2 / p : ℝ))) := by
-  intro v mlpv
+      p p μ μ (A ^ ((2 / p : ℝ))) := fun v mlpv ↦ by
```
grunweg: I think this change actually makes the code less readable. You're switching from term mode immediately to tactic mode, without a need. Can you revert these, please?

Carleson/ToMathlib/HardyLittlewood.lean
```diff
@@ -859,41 +858,34 @@ theorem hasWeakType_maximalFunction_equal_exponents
     rw [maximalFunction_seq_eq _ hp]
     rfl
   let f (k : ℕ) := fun x ↦ maximalFunction μ (tr h𝓑 k) c r (↑p) v x
-  have f_mon : Monotone f := by
-    intro a b hab x
-    apply rpow_le_rpow _ (by positivity)
-    apply iSup₂_le
-    intro i Hi
+  have f_mon : Monotone f := fun a b hab x ↦ by
+    refine rpow_le_rpow (iSup₂_le fun i Hi ↦ ?_) (by positivity)
```
grunweg: Same here; you could move the `intro` into the `refine` call without issues.

Carleson/Psi.lean
```diff
@@ -398,19 +398,6 @@ def C2_1_3 (a : ℕ) : ℝ≥0 := 2 ^ (102 * a ^ 3)
 /-- The constant appearing in part 3 of Lemma 2.1.3. -/
 def D2_1_3 (a : ℕ) : ℝ≥0 := 2 ^ (150 * a ^ 3)
 
-/-- preferably use `kernel_bound` instead. -/
```
grunweg, line 401: This lemma is merely deleted (which is a drive-by change independent of the remaining PR), right?
@fpvandoorn Can you comment on whether this is desirable? (I'm lacking context to do a decision better than "the lemma says don't use and the code compiles".)

Carleson/Psi.lean
```diff
@@ -554,17 +541,16 @@ lemma norm_Ks_le {s : ℤ} {x y : X} :
...
-  · rwa [ENNReal.coe_nnreal_eq, ENNReal.coe_nnreal_eq]
-  · simp only [ne_eq, ENNReal.div_eq_top, ENNReal.ofReal_eq_zero, not_le, ENNReal.ofReal_ne_top,
-      false_and, or_false, not_and]
-    exact fun _ ↦ ne_of_gt (measure_ball_pos volume x (defaultD_pow_pos a s))
+  calc
+    _ ≤ ‖C2_1_3 a / volume.real (ball x (D ^ s))‖ₑ := by
+      rw [← enorm_norm]; exact Real.enorm_le_enorm (norm_nonneg _) norm_Ks_le
+    _ = _ := by
+      rw [div_eq_mul_inv, enorm_mul, enorm_inv]; swap
+      · rw [Measure.real, ENNReal.toReal_ne_zero]; constructor
```
grunweg: Do you need the `constructor` here? ENNReal.toReal_ne_zero generates two side goals, doesn't it?

Carleson/Antichain/TileCorrelation.lean
```diff
@@ -79,10 +79,9 @@ lemma aux_6_2_3 (s₁ s₂ : ℤ) (x₁ x₂ y y' : X) :
   C2_1_3 a / volume (ball x₂ (D ^ s₂)) *
   (D2_1_3 a / volume (ball x₁ (D ^ s₁)) * (edist y y' ^ τ / (D ^ s₁) ^ τ)) := by
   apply mul_le_mul enorm_Ks_le _ (zero_le _) (zero_le _)
-  convert nnnorm_Ks_sub_Ks_le
+  convert enorm_Ks_sub_Ks_le
   rw [← ENNReal.div_rpow_of_nonneg _ _ (τ_nonneg X)]
-  simp only [defaultτ]; congr
-  simp only [coe_nnreal_ennreal_nndist]
+  simp only [defaultτ]
```
grunweg, line 84: Can you un-squeeze this simp, please? (Not required because outside of `ToMathlib`.)

grunweg (commented):
Thanks for doing this. A lot of these changes are clearly for the better.

Some comments: I haven't reviewed the "big files" yet.

Carleson/Antichain/TileCorrelation.lean
Parcly-Taxel: The goal is already solved after `simp only [defaultτ]`.

I have addressed all comments that have been marked as resolved.

Carleson/Psi.lean
fpvandoorn: Yes, this was an old `nnnorm`-variant of the next lemma, so cleaning this up is definitely desired (and fits fine within this PR).

PR #473, opened 2025-07-23
chore: adjust the constants in the blueprint to the formalization

sgouezel:
Also optimize a tiny bit one argument, giving very slightly better constants in the end.

And fix the hyperlinks in the pdf file of the blueprint.

### DIFF
--- a/Carleson/ForestOperator/Forests.lean
+++ b/Carleson/ForestOperator/Forests.lean
@@ -76,7 +76,7 @@ lemma estimate_C7_4_6 {a : ℕ} (n : ℕ) (ha : 4 ≤ a) :
 
 /-- The constant used in `correlation_separated_trees`.
 Has value `2 ^ (512 * a ^ 3 - 4 * n)` in the blueprint. -/
-irreducible_def C7_4_4 (a n : ℕ) : ℝ≥0 := -- SG
+irreducible_def C7_4_4 (a n : ℕ) : ℝ≥0 :=
     2 ^ ((4 * 𝕔 + 12 + 4 * (𝕔 / 4)) * a ^ 3) * 2 ^ (-(4 * n : ℝ))
 
 lemma le_C7_4_4 (ha : 4 ≤ a) : C7_4_5 a n + C7_4_6 a n ≤ C7_4_4 a n := by
@@ -368,7 +368,7 @@ lemma adjointCarlesonRowSum_adjoint
 
 /-- The constant used in `row_bound`.
 Has value `2 ^ (182 * a ^ 3 - n / 2)` in the blueprint. -/
-irreducible_def C7_7_2_1 (a n : ℕ) : ℝ≥0 := -- SG
+irreducible_def C7_7_2_1 (a n : ℕ) : ℝ≥0 :=
   2 ^ ((𝕔 + 7 + 𝕔 / 2 + 𝕔 / 4) * a ^ 3) * 2 ^ (- (n / 2 : ℝ))
 
 lemma le_C7_7_2_1 (a4 : 4 ≤ a) :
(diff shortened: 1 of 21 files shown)


Carleson/ForestOperator/Forests.lean
```diff
@@ -74,9 +74,10 @@ lemma estimate_C7_4_6 {a : ℕ} (n : ℕ) (ha : 4 ≤ a) :
       · convert ih using 1
         ring
 
-/-- The constant used in `correlation_separated_trees`. -/
+/-- The constant used in `correlation_separated_trees`.
+Has value `2 ^ (512 * a ^ 3 - 4 * n)` in the blueprint. -/
 irreducible_def C7_4_4 (a n : ℕ) : ℝ≥0 :=
-    2 ^ ((4 * 𝕔 + 16 + 3 * (𝕔 / 4)) * a ^ 3) * 2 ^ (-(4 * n : ℝ))
+    2 ^ ((4 * 𝕔 + 12 + 4 * (𝕔 / 4)) * a ^ 3) * 2 ^ (-(4 * n : ℝ))
```
grunweg, line 80: You could simplify this further, if you like.

sgouezel: I don't see how. Note that c is a natural number, so 4 * c/4 is not c. In particular for c = 7, then 4 * c/4 is 4, so definitely better than 7.

grunweg: Right, good point.

Carleson/ForestOperator/RemainingTiles.lean
```diff
@@ -807,7 +807,8 @@ lemma cntp_approxOnCube_eq (hu₁ : u₁ ∈ t) :
       rw [Finset.mem_filter_univ, 𝓙₆, mem_inter_iff, mem_Iic] at mp
       rw [indicator_of_mem (mp.2.1 my)]
 
-/-- The constant used in `correlation_near_tree_parts`. -/
+/-- The constant used in `correlation_near_tree_parts`.
+Has value `2 ^ (232 * a ^ 3 - 25/(101a) * Z n κ)` in the blueprint. -/
```
grunweg: Isn't there a factor of `* 2 ^ (21 * a + 5)` missing?

grunweg (commented):
I found one small mistake already; taking a look at the last big file now.

blueprint/src/preamble/common.tex
```diff
@@ -31,10 +30,13 @@
 \newtheorem{example}[theorem]{Example}
 \newtheorem{examples}[theorem]{Examples}
 \numberwithin{section}{chapter}
-\numberwithin{subsection}{section}
-\numberwithin{subsubsection}{subsection}
+%\numberwithin{subsection}{section}
```
grunweg, line 33: Can you elaborate about this change?

blueprint/src/preamble/common.tex
```diff
@@ -31,10 +30,13 @@
...
+%\numberwithin{subsubsection}{subsection}
 \numberwithin{equation}{section}
 
+\usepackage[hidelinks, colorlinks=false, hypertexnames=false]{hyperref}
```
grunweg, line 37: And this one?

grunweg (commented):
I checked the blueprint changes until approximately line 5000, and only spot-checked afterwards. Looks good to me. Thanks for doing all of this!

blueprint/src/preamble/common.tex
sgouezel: Yes. It's just useless, subsections are already numbered within sections by default.

blueprint/src/preamble/common.tex
sgouezel: This fixes the hyperlinks. In the current pdf blueprint, most hyperlinks are broken (try it!), because inner names by hyperref do not include the chapter number. My change tells hyperref to use another strategy to resolve hyperlinks, which works even when there are chapters.

Carleson/ForestOperator/RemainingTiles.lean
sgouezel: Good catch, thanks! Fixed.

PR #475, opened 2025-07-23
chore: cleanup the blueprint pdf file

sgouezel:
Make sure that there is no overfull hbox. Also remove the reference to partial-sum-projection and partial-sum-selfadjoint which have been removed.

After this, there is only one warning left when compiling the blueprint, a reference to a chapter which has not yet been written.

### DIFF
--- a/blueprint/src/chapter/main.tex
+++ b/blueprint/src/chapter/main.tex
@@ -223,7 +223,7 @@ Our concept of a compatible collection $\Mf$ as a natural class of phase functio
 For the proof of \Cref{metric-space-Carleson}, we largely follow \cite{zk-polynomial}, which in turn was inspired by \cite{lie-polynomial}. We make suitable modifications to adapt to our more general setting and have made a few technical improvements in the proof. In particular, in \Cref{overviewsection}, we explicitly divide the main work of the proof into mutually independent sections \ref{thmfromproplinear}, \ref{christsection}, \ref{proptopropprop}, \ref{antichainboundary}, \ref{treesection}, \ref{liphoel}, and \ref{sec-hlm}. Some of these sections follow a similar pattern, starting with a subsection dividing the proof into further mutually independent subsections. This modularization of our proof was strongly endorsed in personal communication by the author of \cite{zk-polynomial}.
 
 \noindent \textit{Acknowledgement.}
-L.B., F.v.D., R.S., and C.T. were funded by the Deutsche Forschungsgemeinschaft (DFG, German Research Foundation) under Germany's Excellence Strategy -- EXC-2047/1 -- 390685813.
+L.B., F.v.D., R.S., and C.T. were funded by the Deutsche For\-schungs\-gemeinschaft (DFG, German Research Foundation) under Germany's Excellence Strategy -- EXC-2047/1 -- 390685813.
 L.B. , R.S., and C.T. were also supported by SFB 1060.
 A.J. is funded by the T\"UBITAK (Scientific and Technological Research Council of T\"urkiye) under Grant Number 123F122.
 
@@ -761,7 +761,7 @@ We also record the following basic estimates for the kernels $K_s$.
     Using $a\ge 4$ proves \eqref{eq-Ks-size}.
 
     To prove \eqref{eq-Ks-smooth} when $2\rho(y,y') > \rho(x,y)$, use the lower bound in
-    \eqref{supp-Ks}, $2\rho(y,y') > \frac{1}{4}D^{s-1}$. Then \eqref{eq-Ks-smooth} follows from
+    \eqref{supp-Ks} and the inequality $2\rho(y,y') > \frac{1}{4}D^{s-1}$. Then \eqref{eq-Ks-smooth} follows from
     the triangle inequality, \eqref{eq-Ks-size} and $a \ge 4$.
 
     If $2\rho(y,y') \le \rho(x,y)$, we rewrite $|K_s(x,y)-K_s(x, y')|$ as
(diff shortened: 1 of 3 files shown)


blueprint/src/preamble/print.tex
```diff
@@ -2,12 +2,18 @@
 % This file is not meant to be built. Build src/web.tex or src/print.text instead.
 
 % We neutralise the Plastex commands
-\newcommand{\uses}[1]{}
 \newcommand{\proves}[1]{}
 \newcommand{\lean}[1]{}
 \newcommand{\leanok}{}
 \tracinglostchars=3
 
+% Make sure that arguments of \uses are real labels, by using invisible refs:
```
grunweg, line 10: Flagging this change for a second pair of eyes.

blueprint/src/chapter/main.tex
```diff
@@ -3805,7 +3805,7 @@ \section{Proof of the Tile Correlation Lemma}\label{sec-tile-operator}
 \Cref{Holder-van-der-Corput} with $z = x_1$ and $R = D^{s_1}$ by \Cref{correlation-kernel-bound}.
 We obtain with $B':= B(x_1, D^{\ps(\fp')})$,
 \begin{equation*}
- {\bf I}(x_1, x_2) \le 2^{8a} \mu(B') \|{\varphi}\|_{C^\tau(B')}|g(x_1)g(x_2)|
+ {\bf I}(x_1, x_2) \le 2^{8a} \mu(B') \|{\varphi}\|_{C^\tau(B')}
```
grunweg, line 3808: Intentional change? Can you elaborate on this?

blueprint/src/chapter/main.tex
```diff
@@ -3010,7 +3010,7 @@ \section{Proof of the Forest Union Lemma}
     Let $\fu' \sim \fu$ be such that $\fp \in \mathfrak{T}_1(\fu')$.
     By the definition of $\sim$, it follows that $\fu' \sim \fu''$.
     By transitivity of $\sim$, we have $\fu \sim \fu''$.
-    By \Cref{relation-geometry}, we have $\sc(\fu'') = \sc(\fu)$, hence $\ps(\fq) < \ps(\fu)$ and $\ps(\fp) \le \ps(\fq) - Z(n+1) \le \ps(\fu) - Z(n+1) - 1$.
+    By \Cref{relation-geometry}, we have $\ps(\fu'') = \ps(\fu)$, hence $\ps(\fq) < \ps(\fu)$ and $\ps(\fp) \le \ps(\fq) - Z(n+1) \le \ps(\fu) - Z(n+1) - 1$.
```
grunweg, line 3013: Can you elaborate on this change? (I cannot find documentation for the `\ps` macro, though I agree `\sc` is bad style.)

blueprint/src/chapter/main.tex
```diff
@@ -223,7 +223,7 @@ \chapter{Introduction}
 For the proof of \Cref{metric-space-Carleson}, we largely follow \cite{zk-polynomial}, which in turn was inspired by \cite{lie-polynomial}. We make suitable modifications to adapt to our more general setting and have made a few technical improvements in the proof. In particular, in \Cref{overviewsection}, we explicitly divide the main work of the proof into mutually independent sections \ref{thmfromproplinear}, \ref{christsection}, \ref{proptopropprop}, \ref{antichainboundary}, \ref{treesection}, \ref{liphoel}, and \ref{sec-hlm}. Some of these sections follow a similar pattern, starting with a subsection dividing the proof into further mutually independent subsections. This modularization of our proof was strongly endorsed in personal communication by the author of \cite{zk-polynomial}.
 
 \noindent \textit{Acknowledgement.}
-L.B., F.v.D., R.S., and C.T. were funded by the Deutsche Forschungsgemeinschaft (DFG, German Research Foundation) under Germany's Excellence Strategy -- EXC-2047/1 -- 390685813.
+L.B., F.v.D., R.S., and C.T. were funded by the Deutsche For\-schungs\-gemeinschaft (DFG, German Research Foundation) under Germany's Excellence Strategy -- EXC-2047/1 -- 390685813.
```
grunweg: Can you do `For\-schungs\-gemein\-schaft`? The last hyphenation opportunity also seems fine to me (as a German native speaker).

blueprint/src/chapter/main.tex
sgouezel: Yes. |g(x_1)g(x_2)| was there on this line and the next one, twice, by mistake, so I removed one.

blueprint/src/chapter/main.tex
sgouezel: \sc means small caps, doesn't work in math mode, so it just didn't do anything here. \ps is what the proof is about, and is needed for the argument to make sense.

blueprint/src/chapter/main.tex
sgouezel: Sure, but the hyphenation happens close to the start of the word. I tried first just with Forschungs\-gemeinschaft, because that would have been really better, but it didn't work out, so I added another possible breakpoint closer to the start of the word. If you tell me that my choices are reasonable, I'm really happy because my German is a little bit rusty :-)

blueprint/src/preamble/print.tex
sgouezel: We had already done this in PFR, for the record, and it helped spot several mistakes.

blueprint/src/chapter/main.tex
grunweg: Yes, your choices are reasonable.
Including another breakpoint seems useful for future robustness --- but since the acknowledgements probably won't be edited much, I won't insist on that.

sgouezel: You are right about robustness. I have added the other breakpoint as you recommended.

PR #478, opened 2025-07-27
Lemma 7.7.2, streamlined

Parcly-Taxel:
This is the tidied-up version of #450. I add a second part to Lemma 7.4.2 to correspond to Lemma 7.7.2 and unify the proof framework for the latter as much as possible. The dependency graph and blueprint are modified to account for this.


Parcly-Taxel:
@grunweg would you please have a look at this?

grunweg:
@fpvandoorn Would you like to? I haven't looked at Lemma 7.7.2 deeply - so me digesting all these details doesn't sound efficient.

fpvandoorn:
Sorry for the long delay. I had time to review this today, and it looks good. Thanks!

PR #482, opened 2025-07-30
Move_defs

mariainesdff:


Carleson/Defs.lean
```diff
@@ -502,609 +214,59 @@ class KernelProofData {X : Type*} (a : outParam ℕ) (K : outParam (X → X →
...
-    _ ≤ ∑ u ∈ 𝓩, (g u).card := Finset.sum_le_sum fun z hz ↦ Finset.card_pos.mpr (g_ne z hz)
-    _ = (𝓩.biUnion g).card := (Finset.card_biUnion (fun z hz z' hz' ↦ g_pd hz hz')).symm
-    _ ≤ 𝓩'.card := by
-      refine Finset.card_le_card fun _ h ↦ ?_
-      rw [Finset.mem_biUnion] at h
-      exact Finset.mem_of_subset (by simp [g]) h.choose_spec.2
-    _ ≤ (2 ^ a) ^ k := c𝓩'
-    _ ≤ _ := by rw [C2_1_1, mul_comm, pow_mul]
+def ClassicalCarleson {f : ℝ → ℂ} (cont_f : Continuous f) (periodic_f : f.Periodic (2 * π)) :
+    Prop := ∀ᵐ x, Filter.Tendsto (S_ · f x) Filter.atTop (nhds (f x))
```
fpvandoorn:
```suggestion
def ClassicalCarleson : Prop :=
  ∀ {{f : ℝ → ℂ}} (cont_f : Continuous f) (periodic_f : f.Periodic (2 * π)),
  ∀ᵐ x, Filter.Tendsto (S_ · f x) Filter.atTop (nhds (f x))
```
Could you change the statement so that it is a single `Prop`, also for the others? Also, could you show in the respective files something like this (so that it's clear the statements are the same)?
```
lemma classical_carleson' : ClassicalCarleson := @classical_carleson

File: Carleson/DoublingMeasure.lean [hunk 2]
```diff
@@ -0,0 +1,555 @@
...
+lemma measurable_Q₁ (x : X) : Measurable (Q x) :=
+  let Q' : X → X → ℝ := fun x' y ↦ Q x' y
+  have : (fun y ↦ Q' x y) = Q x := rfl
+  this ▸ measurable_Q₂.of_uncurry_left
+
+end MeasQ
+
+variable (X) [hK : KernelProofData a K]
+
+include hK
```
Parcly-Taxel, line 551: There are no lemmas that use these `variable` and `include` declarations. These declarations should be removed.

Carleson/Defs.lean
mariainesdff: Done! I called these `classical_carleson_check` etc, because `metric_carleson'` was taken.

PR #486, opened 2025-08-01
add acknowledgement etc

roos-j:
This adds an acknowledgement, fixes an affiliation and fixes some formatting issues in the bibliography.

### DIFF
--- a/blueprint/src/bibliography.bib
+++ b/blueprint/src/bibliography.bib
@@ -262,7 +262,7 @@
 	mrreviewer   = {Joris Roos}
 }
 @article{christ1990b,
-	title        = {A T(b) theorem with remarks on analytic capacity and the Cauchy integral},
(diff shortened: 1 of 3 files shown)


blueprint/src/bibliography.bib
```diff
@@ -902,7 +902,7 @@ @inproceedings{mathlib
 }
 
 @inproceedings{moura2021lean,
-  title = {The Lean 4 Theorem Prover and Programming Language},
+  title = {The {L}ean 4 {T}heorem {P}rover and {P}rogramming {L}anguage},
```
grunweg, line 905: I'm not sure if the capitalisation is required (the [Lean webpage](https://leanprover-community.github.io/lean.bib) capitalises it, but has no special source) --- but won't object either.

fpvandoorn:
Thanks!

PR #492, opened 2025-08-11
refactor `carlesonOn` definition to use `Ks` (task 403)

JasperMS:
The result is closer to the blueprint and the code simpler across the board.


grunweg:
Thanks, looks much nicer indeed!

PR #493, opened 2025-08-14
Cleanup `ProofData`, extract `iHolENorm` and `iLipENorm` (includes task 401)

JasperMS:
While cleaning up `ProofData.lean`, I decided to expand a bit and move the definitions of `iHolENorm` and `iLipENorm` to separate files. Biggest change is to make the tau parameter explicit, IMO this could not be avoided.
There were some unexpected consequences of that because tau is an abbrev and so is unfolded by any `simp` call.

I wasn't sure what ToMathlib folder these should go in, so I put them under the main folder. I played around a bit with the assumptions to try and minimize them.
Probably (quite) some work is required to bring these to Mathlib-worthy completeness, generality and quality, it is expressly not my aim to follow up on that (lacking the specific knowledge to determine appropriate generality etc.).

But at least now the definitions and the results are completely abstracted from the rest of Carleson.


fpvandoorn:
Thanks, this is nice!

We will want a version of these norms in Mathlib, but these definitions are currently sufficiently specialized for this project, that we have to refactor them quite a bit before we do so (as you mention in the file).
But I therefore think these definitions should not be in the `ToMathlib` folder at all, which should consist of things that can be moved to Mathlib without too much changes.

My preference would be: 
* Move the two new files out of `ToMathlib`
* Move the definition `iLipENorm` (but not anything else) to `Defs` so that it remains mostly self-contained with all definitions that occur in the main statement
* Optionally: move files from the top-level `Carleson` to subfolders (`Basic`? Not sure...)

JasperMS:
Done. I didn't create a separate folder because I couldn't think of a good name and a criterion for which files to put there.

fpvandoorn:
Thanks for this, sorry for the delay, and thanks for the ping!

PR #498, opened 2025-08-24
WIP: Proving the HasStrongType upgrade of the main theorem

ldiedering:
This PR works towards proving the upgrade `two_sided_metric_carleson_strong_type` of the main theorem, the key result to derive the L^p version of the classical Carleson-Hunt theorem. 
The chosen approach (see the file `RestrictedWeakType.lean`) uses Lorentz space interpolation. Proving this is a separate project but to be able to apply it here, the main theorem first needs to be upgraded to `hasLorentzType` via `HasRestrictedWeakType.hasLorentzType` (which relies on intermediate results in `Lorentz.lean`). 

Not all proofs are sorry-free but the overall structure has started to converge and hopefully, not too many assumptions are missing anymore.


Carleson/MetricCarleson/Basic.lean
```diff
@@ -397,3 +447,262 @@ lemma enorm_carlesonOperatorIntegrand_le {R₁ R₂ : ℝ≥0} (nf : (‖f ·‖
...
+  unfold carlesonOperator
+  apply Measurable.iSup
+  intro θ
+  --extract_goal
+  apply linearizedCarlesonOperator_measurable hf
+
+
+--generalized from EReal.iSup_add_le_add_iSup
+--TODO: move
+theorem iSup_add_le_add_iSup {ι : Sort*} {β : Type*} [CompleteLattice β] [Add β]
```
fpvandoorn: this is [iSup_add_le](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/CompleteLattice/Group.html#iSup_add_le)

Carleson/MetricCarleson/Basic.lean
```diff
@@ -397,3 +447,262 @@ lemma enorm_carlesonOperatorIntegrand_le {R₁ R₂ : ℝ≥0} (nf : (‖f ·‖
...
+  · apply integrableOn_coi_inner_annulus' _ R₁_pos
+    apply IntegrableOn.mono_set _ (Annulus.oo_subset_ball)
+    apply IntegrableOn.mono_set _ (ball_subset_closedBall)
+    apply hf.integrableOn_isCompact (isCompact_closedBall _ _)
+  · apply integrableOn_coi_inner_annulus' _ R₁_pos
+    apply IntegrableOn.mono_set _ (Annulus.oo_subset_ball)
+    apply IntegrableOn.mono_set _ (ball_subset_closedBall)
+    apply hg.integrableOn_isCompact (isCompact_closedBall _ _)
+
+theorem linearizedCarlesonOperator_add_le_add_linearizedCarlesonOperator
```
fpvandoorn, line 589:
```suggestion
-- TODO: this proof is a good target for automation
theorem linearizedCarlesonOperator_add_le_add_linearizedCarlesonOperator
```

Carleson/MetricCarleson/Basic.lean
```diff
@@ -397,3 +447,262 @@ lemma enorm_carlesonOperatorIntegrand_le {R₁ R₂ : ℝ≥0} (nf : (‖f ·‖
...
+  · exact (integrableOn_coi_inner_annulus' bound_integrable.ofReal hR₁).norm
+  · apply ae_restrict_le
+    apply h_lim.mp
+    apply Eventually.of_forall
+    intro y hy
+    exact Filter.Tendsto.mul (Filter.Tendsto.mul tendsto_const_nhds hy) tendsto_const_nhds
+
+
+--TODO: move
+theorem Filter.iSup_liminf_le_liminf_iSup {α : Type*} {β : Type*} {ι : Sort*} [CompleteLattice α]
```
fpvandoorn: please move to `ToMathlib`

Carleson/ToMathlib/Data/ENNReal.lean
```diff
@@ -145,6 +150,20 @@ lemma biInf_enorm_sub_le {f g : ι → E} :
       add_le_add (by rwa [← tsub_le_iff_left]) (le_biSup (‖g ·‖ₑ) mx)
     _ = _ := by rw [add_rotate]
 
+theorem nonempty_open_inter_Ioi {x : ℝ≥0∞} {u : Set ℝ≥0∞} (isopen_u : IsOpen u) (hx : x ≠ ⊤)
```
fpvandoorn: This should be generalized to `OrderTopology` (or `OrderClosedTopology`), probably with some `DenselyOrdered`, `IsOpen u` should be changed to `u \in \nhds x` and assume `¬ IsMax x`

Carleson/ToMathlib/MeasureTheory/Function/SimpleFunc.lean
```diff
@@ -0,0 +1,369 @@
...
+
+/- Proof stolen from the mathlib SimpleFunc.induction with a minor modification
+   that is even suggested there. -/
+@[elab_as_elim]
+protected theorem induction₀ [MeasurableSpace α] [AddZeroClass γ]
+    {motive : SimpleFunc α γ → Prop}
+    (const : ∀ (c) {s} (hs : MeasurableSet s),
+      motive (SimpleFunc.piecewise s hs (SimpleFunc.const _ c) (SimpleFunc.const _ 0)))
+    (add : ∀ ⦃f : SimpleFunc α γ⦄ (c : γ) ⦃s : Set α⦄ (hs : MeasurableSet s), Disjoint (Function.support ⇑f) s →
+      motive f → motive (SimpleFunc.piecewise s hs (SimpleFunc.const α c) (SimpleFunc.const α 0)) →
```
fpvandoorn, line 29: I think the canonical way to write this function is `(SimpleFunc.const α c).restrict s`. Also, `SimpleFunc.const α 0` can just be written as `0`.

Carleson/ToMathlib/MeasureTheory/Function/SimpleFunc.lean
```diff
@@ -0,0 +1,369 @@
...
+  {a : α} {n : ℕ} :
+    (approx i f n) a ≤ f a := by
+  rw [approx_apply a hf, h_zero]
+  simp only [Finset.sup_le_iff, Finset.mem_range]
+  intro b _
+  split_ifs with hb
+  · exact hb
+  · simp only [bot_le]
+
+theorem iSup_approx_apply' [ConditionallyCompleteLinearOrderBot β] [TopologicalSpace β] [OrderClosedTopology β] [Zero β]
```
fpvandoorn, line 232: Can you try to shorten this proof a bit? For example, use `simp_rw [approx_apply a hf]` before applying `le_antisymm`. Can you `simp_rw` with more tactics before getting 2 goals by applying `le_antisymm`?

Carleson/ToMathlib/MeasureTheory/Function/SimpleFunc.lean
```diff
@@ -0,0 +1,369 @@
...
+/-- A sequence of `ℝ≥0∞`s such that its range is the set of non-negative rational numbers. -/
+def nnrealRatEmbed (n : ℕ) : ℝ≥0 :=
+  Real.toNNReal ((Encodable.decode (α := ℚ) n).getD (0 : ℚ))
+
+theorem nnrealRatEmbed_encode (q : ℚ) :
+    nnrealRatEmbed (Encodable.encode q) = Real.toNNReal q := by
+  rw [nnrealRatEmbed, Encodable.encodek]; rfl
+
+/-- Approximate a function `α → ℝ≥0∞` by a sequence of simple functions. -/
+def nnapprox : (α → ℝ≥0) → ℕ → SimpleFunc α ℝ≥0 :=
```
fpvandoorn, line 278: Why do you need this instead of `eapprox`? Also update docstring.

Carleson/ToMathlib/Order/LiminfLimsup.lean
```diff
@@ -0,0 +1,11 @@
...
+
+theorem Filter.liminf_le_limsup' {α : Type*} {β : Type*} [CompleteLattice α]
+    {f : Filter β} [NeBot f] {u : β → α}
+    --(h : f.IsBoundedUnder (· ≤ ·) u := by isBoundedDefault)
+    --(h' : f.IsBoundedUnder (· ≥ ·) u := by isBoundedDefault)
+    :
+    liminf u f ≤ limsup u f := by
+  apply limsInf_le_limsSup
+  · isBoundedDefault
+  · isBoundedDefault
```
fpvandoorn: I think we don't want this lemma, because it's too close to the existing lemma.
There is a bug with `apply` that `isBoundedDefault` doesn't trigger: https://github.com/leanprover/lean4/issues/10681
Instead of using `apply Filter.liminf_le_limsup'`, you should be able to use `exact Filter.liminf_le_limsup` instead (using the Mathlib-version).

Carleson/ToMathlib/RealInterpolation/Misc.lean
```diff
@@ -1386,6 +1386,15 @@ lemma value_lintegral_res₂ {γ p' : ℝ} {spf : ScaledPowerFunction} (ht : 0 <
   simp [ht', top_rpow_of_pos hσp', mul_top (ENNReal.rpow_pos this spf.hd').ne',
       top_div_of_lt_top ofReal_lt_top]
 
+--TODO: find better name
+lemma wlog_stronglymeasurable {α : Type*} {β : Type*} {mα : MeasurableSpace α} [TopologicalSpace β]
```
fpvandoorn: `AEStronglyMeasurable.induction` or something similar?

Carleson/ToMathlib/WeakType.lean
```diff
@@ -220,6 +220,32 @@ lemma wnorm_toReal_eq {f : α → ℝ≥0∞} {p : ℝ≥0∞} (hf : ∀ᵐ x 
     wnorm (ENNReal.toReal ∘ f) p μ = wnorm f p μ := by
   simp_rw [wnorm, eLpNormEssSup_toReal_eq hf, wnorm'_toReal_eq hf]
 
+/-
+--TODO: move
```
fpvandoorn: remove? Or do you need this later?

Carleson/ToMathlib/MeasureTheory/Function/SimpleFunc.lean
```diff
@@ -0,0 +1,369 @@
+import Mathlib.Algebra.Order.Pi
+import Mathlib.MeasureTheory.Constructions.BorelSpace.Order
+import Mathlib.MeasureTheory.Function.SimpleFunc
+
+noncomputable section
+
+/-
+--TODO: use ConditionallyCompleteLattice instead?
```
fpvandoorn: remove? Or do you need this later?

fpvandoorn (commented):
Thanks, and sorry for the delay in reviewing.
Are the remaining sorry's currently imported in `ClassicalCarleson.lean`? That is not a dealbreaker, but if easy, might be nice to avoid.

Carleson/ToMathlib/Data/ENNReal.lean
ldiedering: I have generalized this now and moved it to `ToMathlib.Topology.Order.Basic`. In addition to your suggestions I needed `LinearOrder` as well.

Carleson/ToMathlib/MeasureTheory/Function/SimpleFunc.lean
ldiedering: As written in the comment above the theorem, I took both the theorem statement (including this notation) and the proof from the mathlib `SimpleFunc.induction` and only made a small modification that is even suggested in a comment to the mathlib version. 
Should I change the notation anyway (and leave a comment that this should be changed in mathlib as well, or just leave it as it is?

Carleson/ToMathlib/MeasureTheory/Function/SimpleFunc.lean
ldiedering: I have tried now but unfortunately one case in the second goal actually needs the original form from before `rw [approx_apply a hf]`. I was still able to slightly shorten the proof by making use of `ciSup_le'` but couldn't find avoidable overlap between the 2 goals.

Carleson/ToMathlib/MeasureTheory/Function/SimpleFunc.lean
ldiedering: It is used in `Measurable.nnreal_induction` and is more natural than `Measurable.ennreal_induction` to apply in `HasRestrictedWeakType.hasLorentzType_helper` because this is about functions to `NNReal`. It wasn't clear to me how to define the motive there for general functions to `ENNReal` and also the induction steps then would have had to be done for general functions to `ENNReal`. 
I've updated the docstrings in this section.

Carleson/ToMathlib/Order/LiminfLimsup.lean
ldiedering: Lemma removed.

Carleson/ToMathlib/MeasureTheory/Function/SimpleFunc.lean
ldiedering: Removed.

Carleson/ToMathlib/WeakType.lean
ldiedering: I think this is essentially the contraposition of `MemWLp.ae_ne_top` so I removed it now.

Carleson/ToMathlib/MeasureTheory/Function/SimpleFunc.lean
fpvandoorn: Try to:
* prove a lemma stating when two suprema are equal, maybe even stated as an iff
* do you really need a bot?

fpvandoorn:
I'll merge this now, this looks good. The last remaining comments can be done in a follow-up PR.

I checked, and you're indeed not importing `sorry` in the main result.

PR #499, opened 2025-08-26
wip: use fun_prop for BoundedCompactSupport

grunweg:
- Many proofs can be golfed nicely. This PR also golfs some other proofs using fun_prop.
- The proof in `AntichainOperator` puzzles me and smells like a fun_prop bug.
- A couple of proofs do not work because `BoundedCompactSupport.indicator` has a measurability side condition; running `fun_prop (discharger := assumption)` did not work. Perhaps a better incantation will fix this.

### DIFF
--- a/Carleson/ForestOperator/Forests.lean
+++ b/Carleson/ForestOperator/Forests.lean
@@ -895,24 +895,21 @@ lemma forest_operator_g_prelude
   calc
     _ = ‖∑ u with u ∈ t, ∫ x, conj (g x) * carlesonSum (t u) f x‖ₑ := by
       congr; rw [← integral_finset_sum]; swap
-      · -- fun_prop fails because of conj vs star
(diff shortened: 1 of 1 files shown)


Carleson/Operators.lean
```diff
@@ -442,17 +447,17 @@ lemma adjointCarlesonSum_adjoint
       unfold carlesonSum; simp_rw [Finset.mul_sum]
     _ = ∑ p with p ∈ ℭ, ∫ x, conj (g x) * carlesonOn p f x := by
       apply integral_finset_sum; intro p _
-      refine hg.conj.mul hf.carlesonOn |>.integrable
+      change Integrable (star g * carlesonOn p f) volume
```
grunweg: This line indicates the issue; that one occurs in many other places as well.

PR #533, opened 2026-01-27
feat: some writing about tactic support and real numbers

grunweg:
Not meant to be merged, but for reference for whoever is interested.


jim-portegies:
I looked at this also in the context of #558 . In my experience, `ring` does work with Ennreals as well, but maybe there are situations in which it doesn't work as well?

There is some information here that I did not incorporate in #558

grunweg:
I remember only anecdotal evidence (some `ring` proofs breaking when moving to `ENNReal`); Maria's comment [here](https://github.com/fpvandoorn/carleson/pull/558#discussion_r3159659013) explains it. `ring` does not work with subtraction, and is weaker when inversion or division are involved (which makes sense).

grunweg:
Let me close this PR in favour of #571.

PR #551, opened 2026-04-09
Lorentz spaces and rearrangement

ldiedering:
The goal was to finally make the results about Lorentz norms sorry-free. Since this has become a pretty lengthy endeavour, here is a summary of the main points:
- Finished the proof of the triangle inequality for Lorentz norms (`eLorentzNorm_add_le`, `eLorentzNorm_add_le'`) and various required lemmas
- Proved the equivalence of the alternative definition of Lorentz norms using decreasing rearrangements (`eLorentzNorm'_eq`) and various required lemmas
- Added many statements and proofs for `rearrangement`
- Moved basics about `distribution`  from `WeakType.lean` to the new file `Distribution.lean`
- Ported the mathlib version of Hölder's inequality to use enorm (`eLpNorm_le_eLpNorm_mul_eLpNorm_of_enorm`) since it was needed in the proof of the triangle inequality
- Ported mathlib results on the layercake formula to use `NNReal` resp. `ENNReal` and `distribution` (see `Layercake.lean`) to use them in `Rearrangement.lean`
- Added new definition `NoAtoms'`, which should eventually replace the current mathlib definition of `NoAtoms` (which is nonstandard)
- Cleanup in files related to Lorentz spaces

Remaining issues: 
- Theory for the new definition `NoAtoms'` needs to be developed but this will be done in a separate project.


fpvandoorn:
Oh sorry, I forgot to mention this in our meeting this morning.

Due to the Mathlib bump, this will probably require one more round of you fixing errors. Hopefully it is not too painful. After that I'll review/merge quickly.

And as discussed, please add upstreaming status notes (can just be a copy-pasted "This file is ready to be upstreamed" if you don't have additional comments).

fpvandoorn:
Thanks!

PR #557, opened 2026-04-27
Paper task 16: Bounded compact support

roos-j:
Here is a first draft. Please feel free to change as needed.
I changed the subsection title to "Test function classes", both to make it less technical sounding and to reflect the fact that we're using two similar classes, `BoundedCompactSupport` and `BoundedFiniteSupport`.

There was a question whether one of the two should be removed and if so which one. I think it might make sense to keep both in this project since this more accurately reflects the blueprint. Upstreaming is another question.
Discussion thread on Zulip: [#Carleson > BoundedCompactSupport vs BoundedFiniteSupport](https://leanprover.zulipchat.com/#narrow/channel/442935-Carleson/topic/BoundedCompactSupport.20vs.20BoundedFiniteSupport/with/591030566)

### DIFF
--- a/paper/main.tex
+++ b/paper/main.tex
@@ -128,7 +128,7 @@ collaborative effort, written and developed in public.
 
 \subsection{Overview of the formalization project}\label{subsec:overview}
 % TODO: I assume that the classical and generalized Carleson theorems will already be mentioned in the previous subsection, so that I can cite them here.
-The goal of the Carleson project was to formalize the proof of the metric space Carleson theorem [ref], proven in \cite{ThieleCarlesonPreprint}, as well as the proof of the classical Carleson theorem [ref] as a corollary of this more general result. 
+The goal of the Carleson project was to formalize the proof of the metric space Carleson theorem [ref], proven in \cite{ThieleCarlesonPreprint}, as well as the proof of the classical Carleson theorem [ref] as a corollary of this more general result.
 
 The original reference \cite{ThieleCarlesonPreprint} was written as a traditional paper for experts in harmonic analysis, but its formalization was since the beginning envisioned as a large scale collaborative project. This posed a problem, since there are not many experts in formalization that are also experts in harmonic analysis, so most of the potential contributors to the formalization would not have the required expertise to fill in the gaps that are considered routine arguments in a harmonic analysis paper.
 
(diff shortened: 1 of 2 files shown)


paper/main.tex
```diff
@@ -307,11 +309,38 @@ \subsection{Working with $L^p$ functions}\label{subsec:Lp}
...
+test function class. We also used the slightly larger class
+of bounded measurable functions supported on a set of finite measure in places where this weaker hypothesis
+matches the blueprint more closely.
+Since most constructions appearing in the proof leave these function classes invariant,
+integrability and measurability considerations are usually mathematically trivial.
+
+It is desirable for this simplicity to be reflected in the formalization.
+This motivated packaging the recurring side conditions into test function classes.
+
+The structures \lean{BoundedCompactSupport} and \lean{BoundedFiniteSupport} record,
```
mariainesdff, line 328: Could you add a short paragraph explaining the conclusion from the Zulip discussion (to keep `BoundedFiniteSupport` in the project, but only upstream `BoundedCompactSupport`)?

paper/main.tex
```diff
@@ -307,11 +309,38 @@ \subsection{Working with $L^p$ functions}\label{subsec:Lp}
...
+respectively, essentially bounded measurable functions with compact support and with support of finite measure.
+Once these hypotheses are bundled, later proofs can invoke short API lemmas instead of repeatedly
+reproving the same measure-theoretic facts. For example, from a hypothesis
+\lean{hf : BoundedCompactSupport f} one immediately obtains that \lean{f} is a member of
+any desired $L^p$ space, % in the context of this project
+or that \lean{s.indicator f} again has bounded compact support when \lean{s} is measurable.
+Lemmas such as \lean{hf.carlesonSum} show that the relevant
+operators preserve the test function class, so once compact support has been established at the input, it remains available throughout the argument.
+
+Further API extensions and better support from \texttt{fun\_prop} could streamline these arguments even more.
```
mariainesdff, line 338: Is there more to say about the use of `fun_prop`?

grunweg: #505 investigated this to some degree. I'd need to remember the status of that, if there were any obstructions to using it more widely or that's just missing a volunteer to try.

paper/main.tex
roos-j: The outcome of the discussion was the we keep both, so no need to change anything here. If upstreaming should be discussed in the paper, I suggest to put this and other upstreaming discussions into an entirely separate section, supporting a possible "impact on mathlib" narrative.

paper/main.tex
roos-j: I have nothing more to say about `fun_prop` at this point. I'd also like be interested to try this later, but maybe we can merge it for now and change later if and when more can be said.

mariainesdff: OK, then let's create a separate task for the `fun_prop` discussion.

PR #561, opened 2026-04-30
add paragraph on standing assumptions in the paper

sgouezel:


paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+tedious repetition in the statements, sometimes at a slight expense in
+readability as the reader has to locate the places where such standing
+assumptions are made. Such a practice is especially important when the
+standing assumptions are lengthy and technical.
+
+The standing assumptions for the statement of the generalization of
+Carleson's theorem are quite formidable. We will not repeat them here,
+see~\cite[Theorem~1.1.1]{CarlesonBlueprint} for details. Let us just mention
+that they involve a metric space $X$, a measure $\mu$ on $X$ which is
+doubling for a constant $2^a$, a family $\Theta$ of functions from $X$ to
```
grunweg, line 324:
```suggestion
a doubling measure for a constant $2^a$, a family $\Theta$ of functions from $X$ to
```
maybe?

paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+$\real$, a distance on $\Theta$ satisfying five properties as well as a
+cancellativity condition, and a kernel $K : X \times X \to \real$ with
```
```suggestion
cancellativity condition, and a kernel $K\colon X \times X \to \real$ with
```

paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+Calderón-Zygmund like properties. The proofs involve even more additional
```
```suggestion
Calderón--Zygmund like properties. The proofs involve even more additional
```
these were two authors

paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+
+One possibility to solve this issue in the formalization would be to say that
+there is no issue, and include as assumptions in each lemma the minimal
+subset of the standing assumptions that is necessary for the statement to
+make sense. This would come at a high readability cost. Moreover, each time
+one applies such a lemma one would need to provide the individual assumptions
+as a long list of parameters to the lemma.
+
+It is a better idea to try to import in the formalization the practice of
+standing assumptions, in some form. One possibility would be to define a
```
```suggestion
standing assumptions, in some form. One option is to define a
```
might be subjective!

paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+structure containing all these standing assumptions, and have it as a
+parameter in every lemma and every definition. The readability cost would be
+minor, but still there. In the Carleson project, we have used another
+strategy, making all the standing assumptions completely implicit -- in
```
```suggestion
strategy, making all the standing assumptions completely implicit --- in
```

paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+particular, we do not need to pass them between lemmas. The implementation
+takes advantage of a feature of the language, typeclasses, initially designed
+for a completely different purpose.
+
+The way to declare that a space is a metric space is the following line:
+\begin{minted}{lean}
+variable {X : Type*} [MetricSpace X]
+\end{minted}
+The declaration between square brackets is the typeclass. When one writes the
+distance \texttt{dist x y} between two points $x$ and $y$ of $X$, the system
```
grunweg, line 354: Does this work?
```suggestion
distance \lean{dist x y} between two points $x$ and $y$ of $X$, the system
```

paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+tries to locate such a metric space typeclass on $X$, and uses the distance
+provided by this typeclass. The same system is used to provide algebraic
+operations on types, and properties of these.
```
grunweg, line 357: Should we add a reference about such uses (such as https://drops.dagstuhl.de/entities/document/10.4230/LIPIcs.ITP.2022.4)?

paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+
+Typeclasses can be used as a device to embody standing assumptions, as
```
grunweg: minor suggestion, might be subjective:
```suggestion
Typeclasses can be used to capture standing assumptions as
```

paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+follows. Let us declare a new typeclass, of the form \texttt{[KernelProofData
```
grunweg, line 360:
```suggestion
follows. Let us declare a new typeclass, of the form \lean{[KernelProofData
```

grunweg (commented):
Thanks a lot for writing this --- clear and informative! I have only very minor comments, some of them subjective (i.e. feel free to disagree with them!).

paper/main.tex
sgouezel: "a measure which is doubling" looks better to me than "a measure which is a doubling measure". No?

paper/main.tex
sgouezel: I prefer the "would be" because of the "would be" in the next sentence, for coherence. Ok for "option" instead of "possibility" if you like it better.

paper/main.tex
sgouezel: It's exactly the same output as far as I can tell, but one is much faster than the other. I think that minted is only useful when there is syntactic coloring.

paper/main.tex
sgouezel: I don't understand your suggestion. Did you mean "to capture standing assumptions as"?

sgouezel (commented):
Thanks for the review! I have commented on the ones that are not obvious to me, let me know what you think.

paper/main.tex
sgouezel: I'm not sure, since it's only basic use.

sgouezel:
-awaiting-author

paper/main.tex
grunweg: Err, indeed (fixed now).

paper/main.tex
grunweg: \lean is a semantic macro (which is nice when we'd want to change the style of all such operations).
If compilation speed is so different, that's a fair argument. I don't feel very strongly.

paper/main.tex
grunweg: ... if "doubling measure" is a property of a measure (as opposed to "doubling measure" being a separate concept), yes. Intuitively, I'd think of "doubling measure" as the latter... what do you think about "a doubling measure on $X$ (for a constant $2^a$)"?
As for the others, I don't feel very strongly here.

paper/main.tex
grunweg: "possibility" repeats with the previous paragraph... so I slightly prefer option.

paper/main.tex
sgouezel: Yes, it's an additional property of a measure (like inner-regular, or nonatomic, or whatever).

paper/main.tex
```diff
@@ -308,7 +308,86 @@ \subsection{Treatment of constants}\label{subsec:constants}
...
+
+One possibility to solve this issue in the formalization would be to say that
+there is no issue, and include as assumptions in each lemma the minimal
+subset of the standing assumptions that is necessary for the statement to
+make sense. This would come at a high readability cost. Moreover, each time
+one applies such a lemma one would need to provide the individual assumptions
+as a long list of parameters to the lemma.
+
+It is a better idea to try to import in the formalization the practice of
+standing assumptions, in some form. One possibility would be to define a
```
sgouezel:
```suggestion
standing assumptions, in some form. One option would be to define a
```

paper/main.tex
sgouezel: Yes, the compilation speed is different by an order of magnitude. In fact I had started with \lean everywhere, and compilation had become unbearably slow, which is why I switched back.

sgouezel:
I'd like to remove the awaiting-author tag, but there's no way I can do it as far as I can tell.

grunweg:
Thanks, LGTM now. Let's wait until Monday to give e.g. Floris a chance to weigh in on the two open questions; if there's no objection or further discussion, I'll be happy to merge as-is.

paper/main.tex
mariainesdff: Does either of you know whether using `listings` instead of `minted` would speed up the compilation time?

paper/main.tex
mariainesdff: I don't think it is needed here.

mariainesdff:
> Thanks, LGTM now. Let's wait until Monday to give e.g. Floris a chance to weigh in on the two open questions; if there's no objection or further discussion, I'll be happy to merge as-is.

I am in favor of merging. Thanks!

paper/main.tex
sgouezel: Yes, listings is quicker. But I think minted is also perfectly acceptable. Since you've already set it up, I think it's fine to stay with minted.

fpvandoorn:
LGTM

PR #565, opened 2026-05-04
Remove the `Countable` assumption from `HardyLittlewood`

lua-vr:
This is a general cleanup towards upstreaming `HardyLittlewood`. In particular it removes the `Countable` assumption from theorems. Some constants also got improved in the process.

I deleted the definition of `MB` in favour of `maximalFunction`, and the proof was restructured to show `measure_biUnion_le_integral` without the `hR` assumption that the radius are bounded first, so that this assumption became obsolete and the results afterwards are achieved more quickly.

- Depends on: #576


lua-vr:
The blueprint is getting out of date with the code with these changes. Should the text and proofs be updated there? What is the policy for the blueprint on these files moving to Mathlib? (I imagine other files would undergo similar blueprint-breaking refactors as well)

Carleson/ToMathlib/HardyLittlewood.lean
```diff
@@ -23,382 +22,227 @@ noncomputable section
...
-    refine AESublinearOn.const (T μ c r i) P (fun hf hg ↦ T.add_le i (hP hf))
-      (fun f d hf ↦ T.smul i) |>.indicator _
+    [IsFiniteMeasureOnCompacts μ] [ProperSpace X] :
+    SublinearOn (maximalFunction (E := E) μ 𝓑 c r 1) (fun f ↦ AEMeasurable f μ) 1 := by
+  refine .iSup₂ fun i hi => .indicator _ ?_
+  simp_rw [inv_one, ENNReal.rpow_one]
+  exact SublinearOn.const (T μ c r i) _ (fun hf hg ↦ by exact T.add_le hf) (fun f d hf ↦ T.smul)
 
 /-- The constant factor in the statement that `M_𝓑` has strong type. -/
-irreducible_def CMB (A p : ℝ≥0) : ℝ≥0 := C_realInterpolation ⊤ 1 ⊤ 1 p 1 (A ^ 2) 1 p⁻¹
```
lua-vr, line 399: What was the benefit of having those constants be irreducible?

Carleson/ToMathlib/HardyLittlewood.lean
```diff
@@ -584,343 +363,41 @@ lemma C_weakType_maximalFunction_lt_top {A p₁ p₂ : ℝ≥0} :
...
-      conv_rhs => enter [2, 2]; rw [← NNReal.rpow_one (_ * _)]
-      gcongr
-      · nth_rw 1 [← mul_one 1]; gcongr
-        · exact (one_le_div (tsub_pos_of_lt hq)).mpr tsub_le_self
-        · norm_cast; rw [← pow_mul]; exact Nat.one_le_two_pow
-      · rw [inv_le_one_iff₀]; right; exact_mod_cast hq.le
-    _ = _ := by ring
-
-/-- Equation (2.0.46). Easy from `hasStrongType_maximalFunction` -/
-theorem hasStrongType_globalMaximalFunction [BorelSpace X] [IsFiniteMeasureOnCompacts μ]
```
lua-vr, line 677: Since the constant is gone, I removed the results specific to `globalMaximalFunction` that became a direct application of some other theorem for `maximalFunction`

Carleson/ToMathlib/HardyLittlewood.lean
```diff
@@ -23,382 +22,227 @@ noncomputable section
...
+      exact MeasureTheory.enorm_ae_le_eLpNormEssSup u μ
+    _ ≤ ⨆ i ∈ 𝓑, (ball (c i) (r i)).indicator (x := x) fun _ ↦ eLpNormEssSup u μ := by
       gcongr; apply setLaverage_const_le
     _ ≤ ⨆ i ∈ 𝓑, eLpNormEssSup u μ := by gcongr; apply indicator_le_self
     _ ≤ eLpNormEssSup u μ := by
       simp_rw [iSup_le_iff, le_refl, implies_true]
 
-protected theorem HasStrongType.MB_top [BorelSpace X] :
-    HasStrongType (fun (u : X → E) (x : X) ↦ MB μ 𝓑 c r u x) ⊤ ⊤ μ μ 1 := by
+-- theorem MeasureTheory.MemLp.maximalFunction_lt_top (hp₁ : 0 < p) (hu : MemLp u ⊤ μ) :
```
lua-vr: This theorem (`maximalFunction_lt_top`) does not seem to be used elsewhere, is there still value in having it?

Carleson/ToMathlib/HardyLittlewood.lean
```diff
@@ -23,382 +22,227 @@ noncomputable section
...
-    ∃ C : Set X, C.Countable ∧ ∀ r > 0, ⋃ c ∈ C, ball c r = univ := by
-  simp_rw [← Metric.dense_iff_iUnion_ball, exists_countable_dense]
-
--- maybe not suited for Mathlib in this form
-lemma countable_globalMaximalFunction :
-    (covering_separable_space X).choose ×ˢ (univ : Set ℤ) |>.Countable :=
-  (covering_separable_space X).choose_spec.1.prod countable_univ
-
--- probably not suited for Mathlib in this form
-lemma exists_ball_subset_ball_two (c : X) {r : ℝ} (hr : 0 < r) :
```
lua-vr, line 43: I just removed those, if there is motivation for keeping them in another file please say so!

Carleson/ToMathlib/HardyLittlewood.lean
```diff
@@ -23,382 +22,227 @@ noncomputable section
...
-/-- The Hardy-Littlewood maximal function w.r.t. a collection of balls 𝓑.
-M_{𝓑, p} in the blueprint. -/
-def maximalFunction (μ : Measure X) (𝓑 : Set ι) (c : ι → X) (r : ι → ℝ)
-    (p : ℝ) (u : X → E) (x : X) : ℝ≥0∞ :=
+variable {X E : Type*} {A : ℝ≥0} [PseudoMetricSpace X] [MeasurableSpace X] [NormedAddCommGroup E]
+  {μ : Measure X} [μ.IsDoubling A]
+  {ι : Type*} {𝓑 : Set ι} {c : ι → X} {r : ι → ℝ} {p : ℝ} {u : X → E} {x : X}
+
+/-- The uncentered Hardy-Littlewood maximal function, for a family of balls. -/
+@[expose] public def maximalFunction (μ : Measure X) (𝓑 : Set ι) (c : ι → X) (r : ι → ℝ) (p : ℝ)
```
lua-vr, line 30: I did not specialize the definition to `p = 1` because it is still applied to a general `p` in `AntichainOperator.lean`

Carleson/ToMathlib/HardyLittlewood.lean
```diff
@@ -23,382 +22,227 @@ noncomputable section
...
+    (u : X → E) (x : X) : ℝ≥0∞ :=
   ⨆ i ∈ 𝓑, (ball (c i) (r i)).indicator (x := x)
     fun _ ↦ (⨍⁻ y in ball (c i) (r i), ‖u y‖ₑ ^ p ∂μ) ^ p⁻¹
 
-/-- The Hardy-Littlewood maximal function w.r.t. a collection of balls 𝓑 with exponent 1.
-M_𝓑 in the blueprint. -/
-abbrev MB (μ : Measure X) (𝓑 : Set ι) (c : ι → X) (r : ι → ℝ) (u : X → E) (x : X) : ℝ≥0∞ :=
-  maximalFunction μ 𝓑 c r 1 u x
+/-- The uncentered Hardy-Littlewood maximal function. -/
+@[expose] public def globalMaximalFunction (μ : Measure X) (p : ℝ) (u : X → E) (x : X) : ℝ≥0∞ :=
```
lua-vr, line 36: But for this one, it could be specialized to `p = 1` without downstream problems if that's convenient

blueprint/src/chapter/main.tex
```diff
@@ -6710,8 +6710,6 @@ \section{Proof of Vitali covering and Hardy--Littlewood}
 The following lemma will be used to define $M$ in the proof of \Cref{Hardy-Littlewood}.
 \begin{lemma}[covering separable space]
     \label{covering-separable-space}
-    \leanok
-    \lean{covering_separable_space}
```
lua-vr, line 6714: -leanok, use mathlib lemma

Carleson/ToMathlib/HardyLittlewood.lean
fpvandoorn: Yes, this seems useful to keep

fpvandoorn (commented):
Thanks, this looks good!
Sorry for the delay in reviewing.

One more point (that I changed my mind about): could you please update the constants in the blueprint to match the formalization? I know it's a pain, but we'd like them to be aligned.

lua-vr:
@fpvandoorn 

> One more point (that I changed my mind about): could you please update the constants in the blueprint to match the formalization? I know it's a pain, but we'd like them to be aligned.

Okay, so I did the following: I manually updated the statements and proofs whose constants were changed in the diff *and* were referenced with `\lean{...}` in the blueprint. I did that visually and by query-replace, because it would take a lot of effort to analyze the proofs. 

Then, I asked a Claude (chat) to independently revise the Lean diffs with the original blueprint, and then diffed the output against my edits and revised that. Its changes turned out to be a superset of what I changed, plus one lemma without the `\lean{...}` tag.

I am not saying this is the proper method and unfortunately I am not sure if anything was left behind, but my opinion is that the constants could be handled better in the next blueprint, at least with macros (and also in the Lean formalization because they are often not defined in terms of the previous ones).

lua-vr:
ps: I think we really should put a warning on the blueprint saying that the constants should be checked in the Lean code. As far as I can tell, it was already the case that the main constant is mismatched with the blueprint!

PR #570, opened 2026-05-05
Write Section 6

mariainesdff:
This is a first draft of the conclusion of the paper, excluding the ongoing related work which will be added by @fpvandoorn later on. It will likely need to be edited once the rest of the paper is completed.

I did not include the project statistics here, since they are already discussed in sections 1.2 and 2.


paper/main.tex
```diff
@@ -521,14 +473,17 @@ \subsection{Common pitfalls}\label{subsec:pitfalls}
...
-      \item Generalize during blueprint writing.
-    \end{itemize}
-\end{itemize}
+%TODO: this will likely need to be revised after \ref{sec:org}, \ref{subsec:rel_work}, \ref{subsec:blueprint_mistakes}, \ref{subsec:blueprint_lessons} and \ref{subsec:pitfalls} are finished.
+
+
+%TODO: The "in Lean" can maybe be removed, but let's wait for \ref{subsec:rel_work} to decide.
+The Carleson project, with a total of 28 contributors, was one of the largest collaborative formalization efforts up to date (excluding extensive mathematical libraries such as \mathlib or MathComp), as well as the first formalization in Lean of a state-of-the-art theorem from the area of harmonic analysis. 
+
+Formalizing the metric Carleson theorem required us to formalize foundational results from harmonic analysis, which we are upstreaming to \mathlib and which will be useful for the formalization of other results from this field. Future project will also benefit from some of the insights we learned during the project, including the creation of extended norms and our reflections about the most convenient ways to work with real number types.
```
grunweg: "upstreaming" is fairly wide-spread in the Lean community, but imho still jargon: "contributing back" or "submitting" perhaps

paper/main.tex
```diff
@@ -521,14 +473,17 @@ \subsection{Common pitfalls}\label{subsec:pitfalls}
...
+
+Beyond topic specific insights, we also learned lessons from the collaborative nature of this project which apply to formalization efforts from any area of mathematics. Our way of assigning and keeping track of ongoing tasks in the Zulip chat was sub-optimal, and if we were to repeat the project, we would instead use the existing GitHub issue infrastructure (as is done for instance in the PNT project). Zulip would still be used as a discussion platform, and in particular to discuss potential errors within the blueprint proofs. To make this more efficient, we would ask one of the topic experts (in our case, an expert in harmonic analysis) to frequently monitor this chat and respond to questions, instead of having F.v.D as an intermediary between the remote formalizers and the harmonic analysis experts.
```
```suggestion
Beyond topic-specific insights, we also learned lessons from the collaborative nature of this project which apply to formalization efforts from any area of mathematics. Our way of assigning and keeping track of ongoing tasks in the Zulip chat was sub-optimal, and if we were to repeat the project, we would instead use the existing GitHub issue infrastructure (as is done for instance in the PNT project). Zulip would still be used as a discussion platform, and in particular to discuss potential errors within the blueprint proofs. To make this more efficient, we would ask one of the topic experts (in our case, an expert in harmonic analysis) to frequently monitor this chat and respond to questions, instead of having F.v.D as an intermediary between the remote formalizers and the harmonic analysis experts.
```

grunweg: Can we link to the PNT project somehow?

grunweg (commented):
LGTM, two minor comments.

We should double-check where the two bullets you moved end up, but this can happen during final editing.

paper/main.tex
mariainesdff: We are already linking to it in Section 2, so I was not sure whether to repeat the link here.

mariainesdff:
> We should double-check where the two bullets you moved end up, but this can happen during final editing.

They were moved to the "Lessons learned" section (I discussed this with Floris, who will be writing it).

grunweg:
Add a TODO about mentioning ongoing related work? Otherwise, let's merge this. "bors d+"

paper/main.tex
```diff
@@ -521,14 +473,17 @@ \subsection{Common pitfalls}\label{subsec:pitfalls}
...
+
+We would also modify some aspects of the blueprint for the formalization. We would use \LaTeX environments for the main definitions in the project, to facilitate their discovery, and we would try to ensure that the results in the blueprint were already proven at the right degree of generality; this would specially apply to results meant for \mathlib, for which we want a general statement and proof, rather than an specialization to the project setting. Finally, we would weight more carefully the benefits of using non-standard proof methods in a hope to simplify the formalization, since this ended up causing most of the mathematical errors in the first version of the blueprint.
```
JasperMS, line 486: Re: "at the right degree of generality". To avoid mismatch issues, it would probably be wise to include both the general statement (and proof) and the specific instance needed.

In some cases, the upstreaming has involved generalizing to a point where the concrete 'application' requires more than just a mere `apply` or `exact`.
Whether to retain those cases where the proof of the concrete instance does reduce to a simple `apply` is of course something that can be judged in the individual instance.

In this way,  we can explicitly serve both Mathlib and the future project without either unduly influencing the other. This can also streamline and thus expedite eventual Mathlib upstreaming.

fpvandoorn:
Let's merge this. I'll suggest some fixes in a follow-up PR.

paper/main.tex
fpvandoorn: I see your point, and I think it depends on the situation. 
* If it is a small/simple lemma, this is probably too much overhead
* If it is a general theorem that requires a nontrivial amount of work to apply in a special case, providing a specialization can is indeed be helpful, especially if that specialization is used more than once.

If a lemma generalizes during the Mathlib PR process such that it cannot be easily applied anymore, then Mathlib itself is sometimes/often already interested in the specialization. If it is not in Mathlib, then (hopefully) that should mean that it's not too bad to apply it in-place.

I don't know if you have specific/compelling examples in Carleson where this happened. If so, feel free to write something and propose a pull request.

PR #580, opened 2026-05-15
L^p version of classical Carleson

ldiedering:
This PR upgrades the proof of classical Carleson to work with the assumption `MemLp f p (volume.restrict (Set.Ioc 0 (2 * π)))` instead of  the stronger `Continuous f` and proves the variant `carleson_hunt` where `f : AddCircle T → ℂ`.


fpvandoorn:
Oh, amazing! That was quick, after finishing the Lorentz space preliminaries.
I guess this still depends on the `sorry`'d Lorentz interpolation? Does it still depend on other sorries? Is it ready to be merged already, or did you intend to make this a draft?

We should keep a sorry-free version of Classical Carleson in this repository. So I think this result should live in a different file until we have it sorry-free. Other than that, I'm happy to live-review on Wednesday and merge it soon after.

ldiedering:
> Oh, amazing! That was quick, after finishing the Lorentz space preliminaries. I guess this still depends on the `sorry`'d Lorentz interpolation? Does it still depend on other sorries? Is it ready to be merged already, or did you intend to make this a draft?
> 
> We should keep a sorry-free version of Classical Carleson in this repository. So I think this result should live in a different file until we have it sorry-free. Other than that, I'm happy to live-review on Wednesday and merge it soon after.

Yes, it depends on sorry'd Lorentz interpolation as well as two sorries that need the new definition of `NoAtoms` but that should be it. It is ready to be merged in principal but indeed at the moment it does not keep a sorry-free path to classical Carleson. But I think it should be relatively quick to just include old versions of some lemmas.

sgouezel:
Can you say where the new definition of ` NoAtoms` is relevant?

ldiedering:
> Can you say where the new definition of ` NoAtoms` is relevant?

It is necessary for `exists_nullmeasurable_between_measure_eq` (in `NoAtoms.lean`) to be true, which is my motivation for changing the definition. This lemma is used in `lintegral_rearrangement_eq` (in `Rearrangement.lean`) which is used to prove `lintegral_rearrangement_add_rearrangement_le_add_lintegral` (here we get rid of the `NoAtoms'` assumption again by taking the product with a nonatomic probability measure; the second sorry is to prove for the new definition that this product measure is nonatomic). This is then used in the proof of the triangle inequality for Lorentz spaces and eventually the proof for the L^p version of Carleson's Theorem depends on it.

sgouezel:
Thanks for the explanation. I agree that we need a name for this notion, but I'm not convinced that it should replace `NoAtoms`. Maybe name it `NoExtendedAtoms`?

ldiedering:
> Thanks for the explanation. I agree that we need a name for this notion, but I'm not convinced that it should replace `NoAtoms`. Maybe name it `NoExtendedAtoms`?

Have you seen this [Zulip discussion](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/MeasureTheory.2ENoAtoms/with/574638121) about this?

sgouezel:
I've posted there, thanks!

Carleson/Classical/Approximation.lean
```diff
@@ -36,6 +39,69 @@ lemma close_smooth_approx_periodic {f : ℝ → ℂ} (unicontf : UniformContinuo
...
+            congr with x
+            unfold f' h
+            rw [AddCircle.liftIoc, AddCircle.liftIoc]
+            simp
+        exact hf'g
+      · apply (eLpNorm_le_of_ae_bound (C := ε') _).trans
+        · simp only [MeasurableSet.univ, Measure.restrict_apply, Set.univ_inter, volume_Ioc,
+          sub_zero]
+          unfold ε'
+          rw [ENNReal.ofReal_rpow_of_pos hT.out,
```
fpvandoorn, line 93:
```suggestion
          -- todo: potentially automatable ENNReal calculation
          rw [ENNReal.ofReal_rpow_of_pos hT.out,
```

Carleson/Classical/Basic.lean
```diff
@@ -55,6 +69,30 @@ lemma partialFourierSum_eq_partialFourierSum' [hT : Fact (0 < 2 * Real.pi)] (N :
     rw [Function.comp_apply, Set.restrict_apply] at this
     exact this
 
+lemma partialFourierSum_eq_partialFourierSum'_apply [hT : Fact (0 < 2 * π)] (N : ℕ) (f : ℝ → ℂ)
```
fpvandoorn: I just noticed that you use `Fact (0 < 2 * π)`, which is not so nice. Much nicer is to have an instance that this is true. You can scope the instance to a namespace `TwoPiPos` or something.

Carleson/Classical/CarlesonHunt.lean
```diff
@@ -12,11 +13,89 @@ noncomputable section
 
 local notation "S_" => partialFourierSum
 
+theorem carleson_hunt_two_pi [hT : Fact (0 < 2 * π)] {f : AddCircle (2 * π) → ℂ} {p : ENNReal}
```
fpvandoorn: Here and elsewhere, please use the (scoped) notation for `ENNReal`, `Real.pi`, `nhds`, ... and open common namespaces like `Filter` to shorten statements.

Carleson/ToMathlib/Analysis/Fourier/AddCircle.lean
```diff
@@ -0,0 +1,29 @@
+module
+
+public import Mathlib.Analysis.Fourier.AddCircle
+public import Carleson.ToMathlib.Topology.Instances.AddCircle.Defs
+
+public section
+
+theorem fourier_comp_equivAddCircle {p q : ℝ} [hp : Fact (0 < p)] [hq : Fact (0 < q)]
+  {x : AddCircle p} {n : ℤ} :
+    (fourier n) ((AddCircle.equivAddCircle p q hp.out.ne' hq.out.ne') x) = (fourier n) x := by
```
fpvandoorn:
```suggestion
    fourier n (AddCircle.equivAddCircle p q hp.out.ne' hq.out.ne' x) = fourier n x := by
```

Carleson/ToMathlib/MeasureTheory/Integral/IntervalIntegral/Periodic.lean
```diff
@@ -0,0 +1,148 @@
...
+
+end Periodic
+
+end Function
+
+namespace AddCircle
+
+open MeasureTheory
+
+lemma volume_preimage_equivIoc {T : ℝ} [hT : Fact (0 < T)] {s : Set ℝ} (hs : MeasurableSet s) :
```
fpvandoorn, line 115: This should (hopefully) be only a few lines using [AddCircle.measurePreserving_mk](https://florisvandoorn.com/carleson/docs/Mathlib/MeasureTheory/Integral/IntervalIntegral/Periodic.html#AddCircle.measurePreserving_mk) and/or [AddCircle.measurePreserving_equivIoc](https://leanprover-community.github.io/mathlib4_docs/Mathlib/MeasureTheory/Integral/IntervalIntegral/Periodic.html#AddCircle.measurePreserving_equivIoc)?

Carleson/ToMathlib/MeasureTheory/Integral/IntervalIntegral/Periodic.lean
```diff
@@ -0,0 +1,148 @@
...
+  --· sorry
+  · rw [eLpNorm'_eq_lintegral_enorm, eLpNorm'_eq_lintegral_enorm, ]
+    congr 1
+    apply setLIntegral_Ioc_add_eq
+    intro x
+    simp only
+    congr 2
+    apply periodic_f
+
+theorem aestronglyMeasurable {t T : ℝ} [hT : Fact (0 < T)] {f : ℝ → ℂ}
```
fpvandoorn, line 61: Add an `iff` version

Carleson/ToMathlib/MeasureTheory/Integral/IntervalIntegral/Periodic.lean
ldiedering: Using `AddCircle.measurePreserving_mk`, I have made it shorter now.

Carleson/ToMathlib/MeasureTheory/Integral/IntervalIntegral/Periodic.lean
ldiedering: Added as `Function.Periodic.aestronglyMeasurable_iff`.

Carleson/Classical/Basic.lean
ldiedering: Done. At some point we should probably try to generalize every part to general periods, avoiding this issue entirely.

ldiedering:
> We should keep a sorry-free version of Classical Carleson in this repository. So I think this result should live in a different file until we have it sorry-free. Other than that, I'm happy to live-review on Wednesday and merge it soon after.

I have added back a sorry-free path to `classical_carleson` now and split and reorganized some files in the `Classical` folder such that the file containing it, `ClassicalCarleson.lean`, should not import any declaration depending on sorry (if I didn't make a mistake; is there an easy way to check?). 

The L^p versions are all in `CarlesonHunt.lean` now, including a proof that they imply `classical_carleson`. Once this file no longer depends on sorry, the files `ClassicalCarleson.lean`, `ControlApproximationEffectContinuous.lean` and `CarlesonOnTheRealLineContinuous.lean` will be obsolete (though for the last one, some of its content might be of independent interest).

fpvandoorn (commented):
Great!

Let me merge this now, but one more comment:
if you write in a basic file
```lean
namespace TwoPiPos
scoped instance : Fact (0 < 2 * π) where
  out := Real.two_pi_pos
end TwoPiPos
```
then in other files you can use the instance by running `open TwoPiPos`.

PR #586, opened 2026-06-22
Add some paragraphs to the conclusion

fpvandoorn:
***Reviewing tip***: look at the diff without the first commit (the first commit only reflows some lines; it only does whitespace changes).

* Some fixes in conclusion
* Write a bit more in the conclusion about Leo's work on Lorentz spaces.

### DIFF
--- a/paper/main.tex
+++ b/paper/main.tex
@@ -1189,7 +1189,8 @@ since this ended up causing most of the mathematical inaccuracies in the first v
 % move this earlier in the conclusion?
 After the Carleson project was finished,
 we have continued to prove further results in the Carleson repository.
-In particular, \Cref{metric-space-Carleson} is strong enough to prove stronger versions of the \Cref{classical-carleson}. In particular, the assumption in \Cref{classical-carleson} that $f$ is continuous can be weakened to the statement that $f$ is $L^p$ for $p>1$. TODO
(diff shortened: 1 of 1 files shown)


paper/main.tex
```diff
@@ -1190,7 +1192,24 @@ \section{Conclusion}\label{sec:conclusion}
...
 we have continued to prove further results in the Carleson repository.
 \Cref{metric-space-Carleson} is strong enough to prove stronger versions of the \Cref{classical-carleson}.
-In particular, the assumption in \Cref{classical-carleson} that $f$ is continuous can be weakened to the statement that $f$ is $L^p$ for $p>1$.
+As Hunt \cite{MR238019} showed in 1968,
+the assumption in \Cref{classical-carleson} that $f$ is continuous
+can be weakened to the statement that $f$ is $L^p$ for $p>1$.
+Proving this requires some additional techniques from analysis.
+
+The \emph{Lorentz-norm} of a function is
+$$\|f\|_{L^{p,q}(X,\mu)}\vcentcolon=
```
grunweg, line 1201: I have the nagging feeling the LaTeX can be uniformized with the rest of paper - but that need not block merging.

grunweg:
All commits but the first one look good to me. Is there a handy way for me to review the first one?

fpvandoorn:
I don't think so... Maybe there is something similar to `git diff --ignore-all-spaces` command that does it, but I think it's fine to leave it unreviewed.

PR #592, opened 2026-06-30
chore: collect key definitions into a basic file

grunweg:
All these definitions are needed to state the main theorems (in a way that only depends on mathlib).
A future PR will depend on this to set up comparator for this repository.


fpvandoorn:
Thanks!
Can you now also minimize the imports of `Carleson/Defs.lean`?

grunweg:
Green, let's merge.

PR #601, opened 2026-07-03
Rewrite the proof roadmap

jim-portegies:
As discussed, a rewrite of the proof roadmap, partly so it isn't a one-to-one copy of the blueprint, but hopefully the new version is a bit easier to understand for an outsider as well.

The permalinks should still be changed but maybe more importantly, there should be some thought put into where to actually include them.

### DIFF
--- a/paper/main.tex
+++ b/paper/main.tex
@@ -539,8 +539,8 @@ and let
     \label{eq-psisum}
     \sum_{s\in \mathbb{Z}} \psi(D^{-s}x)=1
 \end{equation}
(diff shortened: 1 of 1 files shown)


fpvandoorn:
This is good! Let me merge it now, we can refine it later.

I think it would be nice if all major definitions and all formalized theorems/propositions mentioned in the paper get a permalink.

I've used the following in the past to indicate a link, which might be useful to link a whole theorem statement (or multiple parts of a theorem statement). If we use it, we should do so consistently though.
```
\usepackage{fontawesome}
\newcommand{\link}{\ensuremath{{}^\text{\faExternalLink}}}
```

PR #602, opened 2026-07-07
chore: bump to mathlib 4.31

grunweg:
- `positivity` (hence `gcongr`) got a bit stronger
- a bunch of `sdiff` lemmas got renamed to mention `diff`
- a bunch of `convert` had to become `convert!`, a bunch of `simpa using` had to become `simpa using!`
- three files got upstreamed
- mathlib has a new environment linter `defsWithUnderscore`: nolint all exceptions for now (addressing this can and should happen in a follow-up)

PR #603, opened 2026-07-07
ENNReal.lean - add a note

lakesare:
Quick note to make it immediately visible we shouldn't upstream this file.


lakesare:
Closing, Michael is already removing it here https://github.com/fpvandoorn/carleson/pull/602/changes#diff-507434b130af5f2fc2d2c3fd4ac6ec68f75f319b5759e6a98839c0e6780105eb

PR #610, opened 2026-07-14
feat: bump to mathlib 4.32

grunweg:
- a few lemmas got upstreamed
- `NoAtoms` got renamed to `NullSingletonClass` (double-check!)
- the `unusedArguments` linter found a few unused typeclass assumptions
- `Set.le_eq_subset` is now a syntactic tautology; simplify proofs accordingly
- `convert` became intentionally weaker; a few proofs became `convert!` as a result
- remove three unused typeclass hypotheses (found thanks to a fix to the `unusedArguments` linter)

PR #615, opened 2026-07-16
SumEstimates.lean - delete

lakesare:
As mentioned in the comment, it's already upstreamed here https://github.com/leanprover-community/mathlib4/pull/37597.


grunweg:
Thanks, good catch! I'll merge this after #610.

PR #616, opened 2026-07-20
Review Sections 3 through 5.1

mariainesdff:


paper/main.tex
```diff
@@ -1146,13 +1129,12 @@ \subsection{Lessons learned}\label{subsec:blueprint_lessons}
...
 when encountering the smallest or largest scale. This introduced additional formalization effort
 and led to small inaccuracies in the blueprint again causing some confusion during formalization.
-\begin{itemize}
-	\item Refer to general results early on.
-	\item Generalize during blueprint writing.
-\end{itemize}
 
-An additional minor point was that the lack of definition environments in the blueprint also
-meant that definitions were absent from the dependency graph.
+%TODO: double-check by someone who worked on that section
```
grunweg: This is correct, i.e. the TODO can be removed.
```suggestion
```

paper/references.bib
```diff
@@ -219,3 +219,16 @@ @Online{LTEDefinitions
...
+  type = {Conference {{Talk}}},
+  title = {Formalizing invisible mathematics},
+  author = {Bauer, Andrej},
+  year = {2023},
+  month = feb,
+  address = {UCLA Institute for Pure and Applied Mathematics},
+  url = {https://math.andrej.com/2023/02/13/formalizing-invisible-mathematics/},
+  urldate = {2025-09-05},
+  langid = {english},
+  file = {/Users/chambert/Zotero/storage/7QDGK22M/formalizing-invisible-mathematics.html}
```
grunweg: Can you remove this (and check if any other entries in the file contain such paths)?

paper/main.tex
```diff
@@ -1075,39 +1063,34 @@ \subsection{Constant tweaking} As explained below in
...
 
+A more involved fix was required for \cite[Lemma 6.3.4]{CarlesonBlueprint}, whose original proof presented several issues. The first needed change was simple, and consisted on further restricting the allowed values for the parameter $\vartheta$, so that a certain covering property could be applied. However, the proof was also incorrect in the case where the value $s(\mathfrak{p})$ associated to a certain tile $\mathfrak{p}$ took the minimum possible value, and it incorrectly assumed that a certain new tile constructed during the proof would always be distinct from $\mathfrak{p}$. To repair these issues, the statements of both Lemma 6.3.3 and Lemma 6.3.4 had to be modified, and significant changes had to be applied to the proof of Lemma 6.3.4.
```
```suggestion
A more involved fix was required for \cite[Lemma 6.3.4]{CarlesonBlueprint}, whose original proof presented several issues. The first needed change was simple, and consisted of further restricting the allowed values for the parameter $\vartheta$, so that a certain covering property could be applied. However, the proof was also incorrect in the case where the value $s(\mathfrak{p})$ associated to a certain tile $\mathfrak{p}$ took the minimum possible value, and it incorrectly assumed that a certain new tile constructed during the proof would always be distinct from $\mathfrak{p}$. To repair these issues, the statements of both Lemma 6.3.3 and Lemma 6.3.4 had to be modified, and significant changes had to be applied to the proof of Lemma 6.3.4.
```

paper/references.bib
grunweg:
```suggestion
```
I checked that the other entries are fine, so this is enough.

grunweg (commented):
Thanks, only minor comments.

grunweg:
Actually, since my comments we so minor, I've just applied them directly (and will merge the PR now).

PR #617, opened 2026-07-21
feat: related work section

grunweg:


paper/references.bib
```diff
@@ -64,8 +64,6 @@ @misc{vanDoornFormalizedHaarMeasure
 }
 
 
-@Comment{jabref-meta: databaseType:bibtex;}
```
grunweg, line 67: Full disclosure: I'm not sure what this comment means, but since it's also present at the end of the file now, this copy is almost certainly harmless to remove.

paper/main.tex
```diff
@@ -247,13 +247,26 @@ \subsection{Overview of the formalization project}\label{subsec:overview}
 Throughout the article, we will introduce Lean code excerpts from the formalization and use the clickable symbol \faExternalLink\: to link to the corresponding code in our repository or in Mathlib, as appropriate. Some of the code excerpts have been slightly edited for readability.
 
 \subsection{Related work}\label{subsec:rel_work}
-\begin{itemize}
- 	\item Other large-scale formalizations in Lean.
-	\item Other harmonic analysis formalizations.
-\end{itemize}
+
+Large formalization projects have been undertaken in different proof assistants and various areas.
```
```suggestion
Large formalization projects, as measured by the number of authors, have been undertaken in different proof assistants and various areas.
```
Maybe we should make this explicit?

paper/main.tex
```diff
@@ -247,13 +247,26 @@ \subsection{Overview of the formalization project}\label{subsec:overview}
...
+Notable formalization projects in analysis include Immler's verification of the singular hyperbolicity of the Lorentz attractor \cite{ImmlerLorentzSolver} (in Isabelle/HOL),
```
fpvandoorn: This is also single-author. This is not an example of "large", but it could be mentioned as "other formalization in analysis" (though maybe that also doesn't fit, as it's more dynamical systems?)

paper/main.tex
```diff
@@ -247,13 +247,26 @@ \subsection{Overview of the formalization project}\label{subsec:overview}
...
+Immler and Tan's formalization of the Poincaré-Bendixon theorem \cite{ImmlerTanPoincareBendixon} (in Isabelle/HOL), % xxx noteworthy enough? ~7000 LoC
```
fpvandoorn: same as above

paper/main.tex
fpvandoorn: If you don't mean this by "Large" here (since you mention it explicitly next paragraph), then what do you mean here?

paper/main.tex
```diff
@@ -247,13 +247,26 @@ \subsection{Overview of the formalization project}\label{subsec:overview}
...
+the flyspeck project \cite{FlyspeckComplete} formally verifying Hales' proof of the Kepler conjecture (verified using an interplay of results in the Isabelle/HOL and HOL/Light proof assistants), and the formalization of the Prime number theorem in HOL Light \cite{HarrisonPNT} (by Harrison), Isabelle/HOL \cite{AvigadPNT} (by Avigad et al), Metamath \cite{CarneiroPNT} (by Carneiro) and Lean (Kontorovich et al) \footnote{\url{https://alexkontorovich.github.io/PrimeNumberTheoremAnd/}}.
+% skipped four-colour theorem: large, but single person and more combinatorics than analysis or topology
+
+Other mathematical formalization projects involving large collaborations include the formal proof of the Feit--Thompson theorem about finite groups in the Rocq prover \cite{OddOrderTheorem} with a 15 author team, the Liquid Tensor Experiment formalizing an important technical lemma about condensed mathematics in Lean \cites{LTEDefinitions,LTEFinal} (involving 18 authors), the PNT+ project formalising several versions of the Prime Number Theorem using modern tools (with 59 contributors so far) % xxx: is there a better number of authors?
```
fpvandoorn: Stated this way, then foundational libraries such as Mathlib count as well. Maybe write something like "Formalization projects formalizing a single theorem or cluster of theorems"?

paper/main.tex
```diff
@@ -247,13 +247,26 @@ \subsection{Overview of the formalization project}\label{subsec:overview}
...
+A common feature is an explicit mathematical blueprint, a detailed natural language proof carefully written to be accessible for non-experts. In the flyspeck project, this proof was written as a book prior to the formalization \cite{HalesFlyspeckBook}. More recent formalization projects (including the Liquid Tensor Experiment and the Sphere Eversion project \cite{SphereEversionCPP}) have used special software (see Section~\ref{sec:org}) to maintain such a natural language proof side by side with the formal proof.
+
+Prior to the Carleson project, relatively little harmonic analysis had been formalized in any proof assistant: for example, the Isabelle and HOL Light libraries have an implementation of Fourier series \footnote{\url{https://isa-afp.org/entries/Fourier.html}} (but no more advanced tools such as the Hardy-Littlewood maximal function or the real interpolation theorem); the Mizar Mathematical library has a formalization of Minkowski's inequality \footnote{for infinite sums, \url{https://fm.mizar.org/2005-13/pdf13-1/holder_1.pdf}}, but no definition of Fourier series (only discrete Fourier transforms).
```
fpvandoorn: The Hardy-Littlewood maximal function / inequality is formalized before:
https://inria.hal.science/hal-04406350/
But this is done there only for functions from reals to reals:
https://github.com/math-comp/analysis/blob/b647b8f7dc35d68425aa4077344e796706e6024d/theories/lebesgue_integral_theory/lebesgue_integral_differentiation.v#L418

fpvandoorn (commented):
preliminary review

paper/main.tex
grunweg: Well spotted: searching Google scholar for a bit also found this paper. Asking AI more also yielded https://drops.dagstuhl.de/storage/00lipics/lipics-vol309-itp2024/LIPIcs.ITP.2024.5/LIPIcs.ITP.2024.5.pdf by the same authors, but a bit larger. I've added both to our references.

paper/main.tex
grunweg: I'm trying to address both meanings: first one about "notable analysis formalization projects", then "large collaborations". Is the new phrasing clearer?

paper/main.tex
grunweg: Did something like this, good point!

paper/main.tex
grunweg: For me, it is a notable project somewhat related to analysis. I'm fine removing it if we consider it too dissimilar.

(I'm also fine keeping it to illustrate "that's the closest major theorem we could find".)

PR #619, opened 2026-07-22
chore: proofread last part

grunweg:
- proofread Section 5.2 until the end
- Move the first third of Section 5.7 (common pitfalls), about integration on (E)NNReal, to Section 5.3
- Correct the discussion of indicator functions: integrating those is the worse way
- tweak the conclusions section, in light of the related work section (as discussed in-person with Floris)

### DIFF
--- a/paper/main.tex
+++ b/paper/main.tex
@@ -1400,25 +1400,22 @@ since it is better behaved w.r.t.\ multiplication and composition.
 % maybe todo: mention the ICERM group on the integrability tactic?
 
 % \begin{itemize}
-% 	\item \texttt{BoundedCompactSupport} and packaging conditions.
 % 	\item Ongoing experiments with \texttt{fun\_prop}. TODO: consider whether this should be a separate subsection.
 % \end{itemize}
 
 \subsection{Common pitfalls}\label{subsec:pitfalls}
 % TODO: proof-read this section carefully
-As described in \ref{subsec:real}, at the beginning of the project we tried to avoid conversions by using the type \lean{Real} as much as possible, even if quantities were naturally nonnegative or possibly infinite. In the course of the project, in different contexts we realized that using \lean{NNReal} or \lean{ENNReal} instead would lead to shorter proofs and reduce the number of side conditions.
-While generally enabling simplifications, such changes not only came at the cost of additional conversions at the boundary between different contexts but also sometimes necessitated extending the API where \mathlib currently only supports \lean{Real}. For instance, to be able to integrate functions on \lean{NNReal} or \lean{ENNReal}, we added \lean{volume} instances on these types and proved basic properties and relations. For example, when integrating a function $\mlean{f : \ennreal\to \ennreal}$, the following expressions should be equivalent.
+As described in Section~\ref{subsec:real}, at the beginning of the project we tried to avoid conversions by using the type \lean{Real} as much as possible, even if quantities were naturally nonnegative or possibly infinite. In the course of the project, in different contexts we realized that using \lean{NNReal} or \lean{ENNReal} instead would lead to shorter proofs and reduce the number of side conditions.
+While generally enabling simplifications, such changes not only came at the cost of additional conversions at the boundary between different contexts but also sometimes necessitated extending the API where \mathlib currently only supports \lean{Real}. For instance, to be able to integrate functions on \lean{NNReal} or \lean{ENNReal}, we added \lean{volume} instances on these types and proved basic properties and relations. For example, when integrating a function $\mlean{f : \ennreal\to \ennreal}$, the following expressions are mathematically equivalent.
 \begin{lstlisting}[mathescape]
 ∫⁻ (x : ℝ) in Ioi 0, f (ENNReal.ofReal x)
--- TODO: write out the coercion instead!
-∫⁻ (x : ℝ≥0), f (↑x)
+∫⁻ (x : ℝ≥0), f (ENNReal.ofNNReal x)
 ∫⁻ (x : ℝ≥0∞), f x
 \end{lstlisting}
 
 %\begin{minted}{lean}
 %∫⁻ (x : ℝ) in Ioi 0, f (ENNReal.ofReal x)
-%-- TODO: write out the coercion instead!
-%∫⁻ (x : ℝ≥0), f (↑x)
+%∫⁻ (x : ℝ≥0), f (ENNReal.ofNNReal x)
 %∫⁻ (x : ℝ≥0∞), f x
 %\end{minted}
 
--- a/paper/main.tex
+++ b/paper/main.tex
@@ -1341,21 +1341,21 @@ integrability and measurability considerations are usually mathematically trivia
 
 It is desirable for this simplicity to be reflected in the formalization.
 This motivated packaging the recurring side conditions into test function classes.
-
-The structures \lean{BoundedCompactSupport} and \lean{BoundedFiniteSupport} contain essentially bounded measurable functions with compact support and with support of finite measure, respectively. A function $f$ is essentially bounded if there exists a real number $C$ such that $\|f(x)\|\le C$ almost everywhere.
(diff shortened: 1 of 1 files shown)


paper/main.tex
```diff
@@ -1266,17 +1266,16 @@ \subsection{Working with real numbers}\label{subsec:real}
...
+An independently useful, more ambitious yet more powerful tactic would allow reducing goals in $\ennreal$ to $\nnreal$: this could be implemented by case-splitting on each relevant variable being finite or infinite. The finite case allows reduction to $\nnreal$ arithmetic; the infinite case will simplify calculations (as, for example, $\infty$ added to any number in $\ennreal$ is still $\infty$). Such goals can hopefully be proven automatically, or reduced to much simpler tasks for the user.%
+\footnote{For multiplication, as $0 * \infty = 0$, simplifying $a*\infty$ to $\infty$ involves splitting cases by whether $a$ is zero or non-zero. Lest the reader worry about combinatorial explosion, let us remark that (1) splitting on $a$ being zero presumably implies the the case of $a$ being infinite need not be considered separately, and (2) such splits can be performed only as required.}
 The precise design of such a tactic has been discussed (including by some of this paper's authors); its implementation is being worked on.
-% TODO: can I point to something public, perhaps a prototype?
+% TODO: point to a prototype implementation
```
grunweg, line 1276: I have pinged the authors of the ICERM project again, and will hopefully have something to add soon.

paper/main.tex
```diff
@@ -1341,21 +1341,21 @@ \subsection{Test function classes}\label{subsec:bdd_comp_supp}
...
-
-The structures \lean{BoundedCompactSupport} and \lean{BoundedFiniteSupport} contain essentially bounded measurable functions with compact support and with support of finite measure, respectively. A function $f$ is essentially bounded if there exists a real number $C$ such that $\|f(x)\|\le C$ almost everywhere.
+The structures \lean{BoundedCompactSupport} and \lean{BoundedFiniteSupport} contain essentially bounded measurable functions with compact support and with support of finite measure, respectively. A function $f$ is \emph{essentially bounded} if there exists a real number $C$ such that $\|f(x)\|\le C$ almost everywhere.
 
 Once these hypotheses are bundled, later proofs can invoke short API lemmas instead of repeatedly
 reproving the same measure-theoretic facts. For example, from a hypothesis
 \lean{hf : BoundedCompactSupport f} one immediately obtains that \lean{f} is a member of
 any desired $L^p$ space, % in the context of this project
 or that \lean{s.indicator f} again has bounded compact support when \lean{s} is measurable.
+% TODO(MR): I'm not sure if this is clear enough: should we write \lean{BoundedCompactSupport.carlesonSum} instead?
```
grunweg: Flagging this question for broader discussion: I'm not sure if this is clear enough: should we write \lean{BoundedCompactSupport.carlesonSum} instead?

paper/main.tex
```diff
@@ -1370,7 +1370,7 @@ \subsection{Test function classes}\label{subsec:bdd_comp_supp}
 For example, if $g$ is bounded with compact support, and $f$ is a closed embedding,
 is $g\circ f$ a bounded function with compact support?
 This is only true if $g$ has compact support and is bounded everywhere;
-otherwise $f$ may send a set of positive measure to a set of measure 0 where $g$ is ill-behaved.
+otherwise $f$ may send a set of positive measure to a set of measure 0 where $g$ is ill-behaved. % So we have the "worse" definition: is this an issue in Carleson, or just a trade-off to point out?
```
grunweg: Question I got when re-reading this: so the Carleson definition is worse in this specific regard --- is that an issue in practice (if so, how did we deal with it), or simply illustrating a trade-off we made?

paper/main.tex
```diff
@@ -1397,7 +1396,7 @@ \subsection{Test function classes}\label{subsec:bdd_comp_supp}
 since the composition and product of integrable functions need not be integrable.
 The class \lean{BoundedCompactSupport} is a step towards
 making \texttt{fun\_prop} more usable for integrability,
-since it is better behaved w.r.t. multiplication and composition.
+since it is better behaved w.r.t.\ multiplication and composition.
```
grunweg, line 1409: I think the ICERM integrability tactic would fit here. (This would be consistent with the lift_ennreal tactic being mentioned above.)

paper/main.tex
```diff
@@ -1473,15 +1467,14 @@ \section{Conclusion}\label{sec:conclusion}
...
+rather than a specialization to the project setting.
+Finally, we would consider more carefully whether to use non-standard proof methods in a hope to simplify the formalization,
 since this ended up causing most of the mathematical inaccuracies in the first version of the blueprint.
 
 % move this earlier in the conclusion?
 After the Carleson project was finished,
 we have continued to prove further results in the Carleson repository.
-\Cref{metric-space-Carleson} is strong enough to prove stronger versions of the \Cref{classical-carleson}.
+Indeed, \Cref{metric-space-Carleson} is strong enough to prove stronger versions of the \Cref{classical-carleson}.
 As Hunt \cite{MR238019} showed in 1968,
```
```suggestion
Indeed, \Cref{metric-space-Carleson} is strong enough to prove stronger versions of the \Cref{classical-carleson}:
as Hunt \cite{MR238019} showed in 1968,
```
Is this correct, i.e. does "stronger versions" refer to the L^p setting? (If not, can you elaborate on what "stronger versions" should mean?)

paper/main.tex
grunweg: Or would this be correct, for instance (instead of "Proving this requires some additional techniques from analysis")?
```
Proving this can use \Cref{metric-space-Carleson} combined with some additional techniques from analysis.
```

grunweg (commented):
Some comments on particular edits, or locations with questions.

paper/main.tex
grunweg: Discussed in person: leave the writing with dot notation, and put an \extlink to BoundedCompactSupport.carlesonSum.

In general, should use this each time we use a (meaningful) lemma or definition for the first time.

grunweg: For instnace, BoundedCompactSupport and BoundedFiniteSupport should be linked.

paper/main.tex
grunweg: This is illustrating a downside of the current trade-off. I will see if the phrasing can be tweaked.

Maybe this should be merged with the subsequent paragraph.

paper/main.tex
grunweg: Discussed: maybe have a brief discussion (but keep it short, since the authors there are mostly disjoint with the authors here).

paper/main.tex
grunweg: Discussed: would be better. In fact, this uses real Carleson, so referencing that instead would be even better.

```
The \emph{Lorentz norm} of a function is
```
(apply in the right line)

paper/main.tex
grunweg: Upon second thought, I'm not sure. I've tweaked the TeX comment with a potential suggestion, but will leave it as-is for now.

paper/main.tex
grunweg: I've said something. I think this is not ideal, but good enough.

paper/main.tex
grunweg: Addressed the particular comment; the general item is tracked in review items.

paper/main.tex
```diff
@@ -1266,17 +1266,28 @@ \subsection{Working with real numbers}\label{subsec:real}
...
 The precise design of such a tactic has been discussed (including by some of this paper's authors); its implementation is being worked on.
-% TODO: can I point to something public, perhaps a prototype?
+% TODO: point to a prototype implementation
 
-As a final example around considerations on which number types to use,
-consider the supremum of a set.
+The choice of real number type also came up when integrating over a set. Before the Carleson project, there was no definition of integral of functions on \lean{NNReal} or \lean{ENNReal}. As integrating such functions was useful, we added \lean{volume} instances on these types and proved basic properties and relations. For example, when integrating a function $\mlean{f : \ennreal\to \ennreal}$, the following expressions are mathematically equivalent. This avoids another source of casting between real number types.
```
```suggestion
The choice of real number type also came up when integrating over a subset of real numbers. Before the Carleson project, there was no definition of integral of functions on \lean{NNReal} or \lean{ENNReal}. As integrating such functions was useful, we added \lean{volume} instances on these types and proved basic properties and relations. For example, when integrating a function $\mlean{f : \ennreal\to \ennreal}$, the following expressions are mathematically equivalent. This avoids another source of casting between real number types.
```

PR #622, opened 2026-07-24
fix: make LaTeX code compile

grunweg:


grunweg:
The paper builds fine in CI now (which was not the case before): let me merge this now.

PR #624, opened 2026-07-24
Basics for `NoAtoms`

ldiedering:
This PR adds more basics for `NoAtoms`, sufficient to fill in all related sorries in the project. 
Main results are `NoAtoms'.exists_measurable_between_measure_eq`, `NoAtoms'.prod.instNoAtoms_fst` (product measures inherit the `NoAtoms'` property if both measures are sigma finite) and `NoAtoms'.of_metric`. 
I had to add sigma finiteness assumptions in some other places (notably the triangle inequality for Lorentz spaces). Maybe there is a way to circumvent this but at the moment it does not hurt since all usages fulfill the assumption.


fpvandoorn:
Oh great, that is very quick! `SigmaFinite` is no problem.

PR #625, opened 2026-07-25
chore(paper): proofreading items 1-4

pitmonticone:
Items 1-4 from the [proofreading thread](https://leanprover.zulipchat.com/#narrow/channel/442935-Carleson/topic/Proofreading.20the.20paper/near/612631089):

- use `\subseteq` for inclusion of sets, and `\supseteq` for the tile order
- use `\le` for the order on cubes, matching `Grid.le_def` and `le_topCube`
- refresh the external links whose line numbers had gone stale, and add 24 new ones
- remove 21 equation labels that are never referenced
- escape the fragment as `\#` in the `Finset.Icc` footnote, where hyperref needs it


grunweg:
Thanks for the PR! I just reviewed the diff until line 400 (inclusive) and have no changes to suggest.

While I appreciate the desire to standardize all external links to one commit (and agree we should do this in the end), I don't think it's time to do this for Carleson just yet. For example, #631 will change the statements of the maximal function again (and #633 update the paper accordingly), so the code referenced should include this change. (Another reason might be upstreaming: if anything we're linking to gets upstreamed to mathlib in the meantime, we should do a final check for this.)

Can you revert the trivial updates to Carleson links after 400? (Changes changing line numbers are fine; everything above line 400 also is. But reviewing the others is just unnecessary, given that we'll bump the commit anyway.) Thanks!

pitmonticone:
I see, that makes perfect sense. I'll address this later today or tomorrow. Thanks!

pitmonticone:
Done.

paper/main.tex
```diff
@@ -511,7 +508,7 @@ \subsubsection{Metric space Carleson}\label{subsubsec:metric-carleson}
 
 We prove three versions of the main result.
 The first version is the following restricted weak type estimate for $T$ in the range $1<q\le 2$,
-which by interpolation techniques recovers $L^q$ estimates for the open range $1<q<2$.
+which by interpolation techniques recovers $L^q$ estimates for the open range $1<q<2$.\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/MetricCarleson/Main.lean#L203}{\extlink}
```
grunweg: Can you move this into the theorem reference? If you escape the #fragment, that should not be a problem.

paper/main.tex
```diff
@@ -534,7 +530,7 @@ \subsubsection{Metric space Carleson}\label{subsubsec:metric-carleson}
 The fact that we don't assume H\"older-continuity of the kernel $K$ in the first argument allows us to also capture this situation.
 
 The second version of our main result has a slightly weaker replacement of the assumption \eqref{nontanbound}.
-It implies the Walsh-case of Carleson's theorem.
+It implies the Walsh-case of Carleson's theorem.\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/MetricCarleson/Linearized.lean#L109}{\extlink}
```
grunweg: Idem

paper/main.tex
```diff
@@ -550,27 +546,25 @@ \subsubsection{Metric space Carleson}\label{subsubsec:metric-carleson}
...
     $|f|\le \mathbf{1}_F$, we have, with $T_\tQ$ defined in \eqref{def-lin-main-op},
     \begin{equation}
-        \label{linresweak}
         \left|\int_{G} T_\tQ f \, \mathrm{d}\mu\right| \le \frac{2^{443a^3}}{(q-1)^6} \mu(G)^{1-\frac{1}{q}} \mu(F)^{\frac{1}{q}}\, .
     \end{equation}
 \end{theorem}
 
 The third version of the main result involves two-sided Calder\'on--Zygmund kernels, whose additional regularity allows us to weaken one of the assumptions to the family of operators $T_r$, which is easier to work with in applications.
-In particular, the two-sided case can be applied to derive the classical Carleson's theorem.
+In particular, the two-sided case can be applied to derive the classical Carleson's theorem.\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/TwoSidedCarleson/MainTheorem.lean#L30}{\extlink}
```
grunweg: And here

paper/main.tex
```diff
@@ -616,26 +608,25 @@ \subsubsection{Proof roadmap}\label{subsubsec:proof-sketch}
...
-\begin{equation}\label{coverdyadic}
-    I\subset \bigcup_{J\in \mathcal {D}: s(J)=k}J\, .
+\begin{equation}
+    I\subseteq \bigcup_{J\in \mathcal {D}: s(J)=k}J\, .
 \end{equation}
 \item \label{it:dyadicproperty} For every non-disjoint dyadic cubes $I, J$ with $s(I) \leq s(J)$, we have that
-\begin{equation}\label{dyadicproperty}
-    I\subset J.
+\begin{equation}
+    I\le J.
```
grunweg, line 617: I think this should be `\subseteq`; compare https://github.com/fpvandoorn/carleson/blob/17645d176450a32d37adbb00a4198e7d5127e301/Carleson/GridStructure.lean#L38

Is there anything I'm missing?

paper/main.tex
```diff
@@ -616,26 +608,25 @@ \subsubsection{Proof roadmap}\label{subsubsec:proof-sketch}
...
 \end{equation}
 \item \label{it:subsetmaxcube} The exists a cube $I_0 \in \mathcal{D}$ with scale $s(I_0) = S$ and center $c(I_0) = o$ (the point where all $\mfa \in \Mf$ vanish), and for all dyadic cubes $J$ it holds that
 \begin{equation}\label{subsetmaxcube}
-    J \subset I_0\,.
+    J \le I_0\,.
```
grunweg, line 621: Same question here.

paper/main.tex
grunweg: Ah, per `Grid.le_def` this is equivalent in this case. So the question is which of these is more conceptual. What do you think?

paper/main.tex
```diff
@@ -651,28 +642,28 @@ \subsubsection{Proof roadmap}\label{subsubsec:proof-sketch}
 % is injective.
 \item \label{it:omega-disjoint} For every dyadic cube $I$ and every pair of tiles $\fp \neq \fq$ in $\scI^{-1}(\{I\})$, the sets $\fc(\fp)$ and $\fc(\fq)$ are disjoint.
 \item \label{it:omega-cover} For every dyadic cube $I$, we have that
-\begin{equation}\label{eq-dis-freq-cover}
```
grunweg, line 655: Note to self: still need to review the next ten lines.

paper/main.tex
```diff
@@ -651,28 +642,28 @@ \subsubsection{Proof roadmap}\label{subsubsec:proof-sketch}
...
 \end{equation}
 with
-\begin{equation}\label{defdp}
+\begin{equation}
     d_{\fp} := d_{B(\pc(\fp),\frac 14 D^{\ps(\fp)})}\, .
 \end{equation}
 \end{enumerate}
 
-With the above definitions, it can be shown that the finitary Carleson theorem reduces to the discrete Carleson theorem.\href{https://github.com/fpvandoorn/carleson/blob/3e96104fa4cd6d5ff37434ac39c49604c86a7658/Carleson/Discrete/MainTheorem.lean#L45}{\extlink}
+With the above definitions, it can be shown that the finitary Carleson theorem reduces to the discrete Carleson theorem.\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/Discrete/MainTheorem.lean#L44}{\extlink}
```
grunweg: Same comment as above: can you move the code link into the theorem, please?

paper/main.tex
```diff
@@ -711,7 +701,7 @@ \subsubsection{Proof roadmap}\label{subsubsec:proof-sketch}
 and we classify tiles by how many marked tiles are ``near'' a given tile.
 Let us note that these characterizations also take into account the orders on cubes and tiles.
 The order on tiles is given by $\fp\le \fp'$ iff
-$\scI(\fp)\le \scI(\fp')$ and $\Omega(\fp)\supset \Omega(\fp')$.
+$\scI(\fp)\le \scI(\fp')$ and $\Omega(\fp)\supseteq \Omega(\fp')$.
```
grunweg, line 705: Note to self: double-check this line

paper/main.tex
```diff
@@ -741,16 +731,16 @@ \subsection{Stating the main results in Lean}\label{subsec:lean-statements}
...
...prover-community/mathlib4/blob/81a5d257c8e410db227a6665ed08f64fea08e997/Mathlib/Order/Interval/Finset/Defs.lean#L280}{\extlink}\footnote{In any preorder $\alpha$, one can define intervals which on each side can be either open (o), closed (c) or infinite (i). For instance, \lean{Set.Ico a b}\href{https://github.com/leanprover-community/mathlib4/blob/81a5d257c8e410db227a6665ed08f64fea08e997/Mathlib/Order/Interval/Set/Defs.lean\#L59}{\extlink} denotes the inverval $[a, b)$ and \lean{Set.Iic b}\href{https://github.com/leanprover-community/mathlib4/blob/81a5d257c8e410db227a6665ed08f64fea08e997/Mathlib/Order/Interval/Set/Defs.lean\#L44}{\extlink} denotes $(-\infty, b]$. If moreover $\alpha$ is a \lean{LocallyFiniteOrder}\href{https://github.com/leanprover-community/mathlib4/blob/81a5d257c8e410db227a6665ed08f64fea08e997/Mathlib/Order/Interval/Finset/Defs.lean\#L108}{\extlink} (i.e., all bounded intervals in $\alpha$ are finite), \lean{Finset.Icc, Finset.Ico, Finset.Ioc} and \lean{Finset.Ioo} provide \lean{Finset}\href{https://github.com/leanprover-community/mathlib4/blob/81a5d257c8e410db227a6665ed08f64fea08e997/Mathlib/Data/Finset/Defs.lean\#L72-L79}{\extlink} versions of these intervals.
```
grunweg: note to self: compare!

paper/main.tex
```diff
@@ -785,7 +775,7 @@ \subsection{Stating the main results in Lean}\label{subsec:lean-statements}
   (h : ball x₁ r₁ ⊆ ball x₂ r₂) : dist_{x₁, r₁} f g ≤ dist_{x₂, r₂} f g
 \end{lstlisting}
 
-The class \lean{IsOneSidedKernel}\href{https://github.com/fpvandoorn/carleson/blob/f2f614db47f1a30e3fceaa9c2e1f99ff356dd577/Carleson/Defs.lean#L240}{\extlink} captures the notion of one-sided Calder\'on--Zygmund kernel (Definition~\ref{def:one-sided-kernel}).
+The class \lean{IsOneSidedKernel}\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/Defs.lean#L250}{\extlink} captures the notion of one-sided Calder\'on--Zygmund kernel (Definition~\ref{def:one-sided-kernel}).
```
grunweg: Is this the second time this class is linked? Once would probably suffice.

paper/main.tex
```diff
@@ -805,7 +795,7 @@ \subsection{Stating the main results in Lean}\label{subsec:lean-statements}
   hcz : IsOneSidedKernel a K
 \end{lstlisting}
 
-The statement of the metric Carleson theorem also requires fixing two positive real numbers $q, q'$ satisfying the equality $q^{-1} + q'^{-1} = 1$, encoded by definition \lean{HolderConjugate}\href{https://github.com/leanprover-community/mathlib4/blob/5a0753dc725610d074b177e4677000ee96165b8e/Mathlib/Data/Real/ConjExponents.lean#L59}{\extlink} from \mathlib.
+The statement of the metric Carleson theorem also requires fixing two positive real numbers $q, q'$ satisfying the equality $q^{-1} + q'^{-1} = 1$, encoded by definition \lean{HolderConjugate}\href{https://github.com/leanprover-community/mathlib4/blob/81a5d257c8e410db227a6665ed08f64fea08e997/Mathlib/Data/Real/ConjExponents.lean#L59}{\extlink} from \mathlib.
```
grunweg: Revert, please?

grunweg (commented):
I've reviewed a bit further, until line 830 (with a few things I should double-check later). More minor comments only.

grunweg:
Can you also update the PR description to the current reality?
> pin all external links to one commit per repository, and add 25 more

is no longer accurate

paper/main.tex
pitmonticone: `\le`, I'd say. It's the form the rest of the dev uses, and `relative_fundamental_dyadic` takes `𝓘 p ≤ 𝓘 p'` as its hypothesis, so `\subseteq` here would mean switching notation for the same relation a few conditions later which... isn't ideal IMO. 

But I think your objection is fair because it was used long before being explained, so I've now defined it where the grid structure is introduced.

WDYT?

paper/main.tex
pitmonticone: Same answer.

pitmonticone:
Thank you @grunweg for your review! I should have addressed all your review comments.

paper/main.tex
fpvandoorn: Yes, sounds good.

fpvandoorn (commented):
Thanks for this! Let's get this is, so that others can see the newest version.

One more comment: can you use the `equation*` environment for all the unreferenced equations? Then the label also doesn't appear in the pdf.

PR #629, opened 2026-07-27
refactor: remove `ENormedSpace`

grunweg:
The current definition means `[ENormedSpace E]` is equivalent to `[ENormedAddCommMonoid E] [Module ℝ≥0 E] [ENormSMulClass ℝ≥0 E]`: remove this composite definition (and replace the instance by instances of the respective classes).
In the process, slightly weaken the typeclass assumptions on a few lemmas.


Carleson/ToMathlib/ENorm.lean
```diff
@@ -31,7 +31,9 @@ class ENormedAddCommSubMonoid (E : Type*) [TopologicalSpace E] extends ENormedAd
   esub_self : ∀ x : E, x - x = 0
 
 /-- An enormed space is an additive monoid endowed with a continuous enorm.
-Note: not sure if this is the "right" class to add to Mathlib. -/
+Note: not sure if this is the "right" class to add to Mathlib.
+In fact, it is now just the conjunction of the classes `ENormedAddCommMonoid E`, `Module ℝ≥0 E` and
+`ENormSMulClass ℝ≥0 E`: should this be deprecated and deleted instead? -/
 class ENormedSpace (E : Type*) [TopologicalSpace E] extends ENormedAddCommMonoid E, Module ℝ≥0 E where
   enorm_smul_eq_smul : ∀ (c : ℝ≥0) (x : E), ‖c • x‖ₑ = c • ‖x‖ₑ
```
fpvandoorn: You can just remove this definition if it's not used in Carleson anymore (no need to deprecate, since we don't have downstream users).

You can optionally add a comment "the generalization of `NormedSpace` to `ENorm` is `[ENormedAddCommMonoid E] [Module ℝ≥0 E] [ENormSMulClass ℝ≥0 E]`"

fpvandoorn:
LGTM, feel free to merge

Carleson/ToMathlib/ENorm.lean
grunweg: Good idea, done (and kept the instances below we were still missing).

PR #630, opened 2026-07-27
chore(LorentzSeminorm/Basic): weaken lemma hypotheses, re-order for better variable management

grunweg:
Weaken the lemma hypotheses to the minimal assumptions needed: for many lemmas, that is just an enorm or a continuous enorm. Reorder lemmas according to their minimal hypotheses (which is actually also quite a coherent grouping).
While at it, correct some obvious indentation errors.


grunweg:
@ldiedering Just pinging you so you're aware. This should not conflict with your work, but there might be merge conflicts.

PR #633, opened 2026-07-27
feat: mention that our definition of Hardy-Littlewood maximal functio…

grunweg:
…n is more general than the literature.


paper/main.tex
```diff
@@ -1042,10 +1042,14 @@ \subsubsection{The Hardy--Littlewood maximal function} An important ingredient i
...
 
-For the formal proof, we changed the definition to allow arbitrary suprema: this means the maximal function is $[0,\infty]$-valued. Being able to speak of $L^p$-functions valued in $[0,\infty]$ prompted the introduction of \emph{extended norms}, see \S\ref{subsec:enorm}.
+For the formal proof, we changed the definition to allow arbitrary suprema: this means the maximal function is $[0,\infty]$-valued.
+(Our definition also extends to a far broader class than functions $\complex\to\real^d$. Initially, we allowed any function $X\to E$ from a pseudo-metric space $X$ with a measure, to a normed abelian additive group $E$.
+The final form\href{https://github.com/fpvandoorn/carleson/blob/696c07687a9c26c87c8a1d36b16da3775d9d11e8/Carleson/ToMathlib/HardyLittlewood.lean#L32}{\extlink}
+allows $E$ to be any space with an extended norm, requiring no additive structure.)
+Being able to speak of $L^p$-functions valued in $[0,\infty]$ prompted the introduction of \emph{extended norms}, see \S\ref{subsec:enorm}.
```
fpvandoorn, line 1051: I don't think we should have a 3-sentence parenthetical. Also, I think the first and the last sentence of this paragraph below together. I would write something like this:

> For the formal proof, we changed the definition to allow arbitrary suprema: this means the maximal function is $[0,\infty]$-valued. Being able to speak of $L^p$-functions valued in $[0,\infty]$ prompted the introduction of \emph{extended norms}, see \S\ref{subsec:enorm}.
We also generalized the allowed input functions to any function from a pseudo-metric space to be any space with an extended norm.

grunweg: Sure; reworded (but kept the link to the definition).

PR #635, opened 2026-07-28
chore: bump to current mathlib

grunweg:
- lots of deprecation warnings around `setOf` -> `ofPred` renamings
- a bunch of deprecation warnings around `Set.restrict` renamed to `Set.domRestrict`
- a bunch of deprecation options `backward.isDefEq.respectTransparency.types` are now necessary (and indicate pre-existing defeq abuse)
- one file has been upstreamed to mathlib

PR #636, opened 2026-07-28
chore: bump to current mathlib (continuing)

lakesare:
- Fix the sorry in `LargeSeparation.lean` (by replacing `lia` with `omega`).
- Remove `ToMathlib/Order/Interval/Set/Disjiont.lean`: the only theorem left it in is already in mathlib (`as IsGLB.biUnion_Ioi_eq`)

Continues #635.


grunweg:
Thanks. I condensed the PR description a bit, otherwise looks great!

PR #637, opened 2026-08-05
chore: fix the build

grunweg:
#624 had not been updated for the mathlib bump; fix the few deprecations and newly necessary backward compatibility options by hand. Also insert a few upstreaming comments.


fpvandoorn:
Thanks!

PR #638, opened 2026-08-06
chore: paper proofreading

js2357:
Mostly minor spelling/grammar/typo corrections. The most interesting ones:

- The definition of "grid structure" had been missing some details (such as the definition of the function s). Besides adding the missing details, I moved the definition of "dyadic cube" into its own paragraph so that the paragraph defining "grid structure" doesn't need to do quite so much work.
- The definition of "grid structure" relies on the constant kappa, which was never mentioned prior to its use there. I placed the definition of kappa next to the definition of D since they're both constants defined off the constant a. Unfortunately, this introduces kappa a few paragraphs before it's used, but I don't think that's too bad.
- Previously, the function e(r) = e^{ir} was defined the second and third times it was used, but not the first time. I added the definition to the first use. I didn't remove either of the later definitions; do we think it's overkill to define the notation three times?


paper/main.tex
```diff
@@ -597,26 +601,26 @@ \subsubsection{Proof roadmap}\label{subsubsec:proof-sketch}
...
+A grid structure consists of dyadic cubes. A \emph{dyadic cube} is a pair $(I, k)$ of a Borel set $I$ in $X$ and an integer $k \in [-S, S]$ for some fixed natural number $S$. Usually we abuse notation and write $I$ for a dyadic cube $(I, k)$. We write $I \le J$ to mean that for two cubes $(I, k)$ and $(J, l)$, we have $I \subseteq J$ and $k \le l$.
 
-The conditions that the objects in a grid structure satisfy are as follows.
+A \emph{grid structure}\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/GridStructure.lean#L20}{\extlink} on $X$ is a triple $(\mathcal{D}, c, s)$ consisting of a finite set $\mathcal{D}$ of dyadic cubes, a function $c\colon \mathcal{D} \to X$ called the center function, and a surjective function $s\colon \mathcal{D} \to [-S, S]$, $(I, k) \mapsto k$, called the scale function, that together satisfy the conditions \ref{it:coverdyadic}-\ref{it:eq-small-boundary}:
```
```suggestion
A \emph{grid structure}\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/GridStructure.lean#L20}{\extlink} on $X$ is a triple $(\mathcal{D}, c, s)$ consisting of a finite set $\mathcal{D}$ of dyadic cubes, a function $c\colon \mathcal{D} \to X$ called the center function, and a surjective function $s\colon \mathcal{D} \to [-S, S]$, $(I, k) \mapsto k$, called the scale function, that together satisfy the conditions \ref{it:coverdyadic}--\ref{it:eq-small-boundary}:
```
maybe?

paper/main.tex
```diff
@@ -597,26 +601,26 @@ \subsubsection{Proof roadmap}\label{subsubsec:proof-sketch}
...
 
-A \emph{grid structure}\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/GridStructure.lean#L20}{\extlink} on $X$ is a triple $(\mathcal{D}, c, s)$ consisting of a finite set $\mathcal{D}$ of so-called dyadic cubes, a function $c\colon \mathcal{D} \to X$ called the center function, and a surjective function $s\colon \mathcal{D} \to [-S, S]$ called the scale function, that together satisfy the conditions \ref{it:coverdyadic}-\ref{it:eq-small-boundary} below. Here, a dyadic cube is just a pair $(I, k)$ of a Borel set $I$ in $X$ and an integer $k \in [-S, S]$. Usually we abuse notation and write $I$ for a dyadic cube $(I, k)$. The dyadic cubes are ordered by $I \le J$ if and only if $I \subseteq J$ and $s(I) \le s(J)$.
+A grid structure consists of dyadic cubes. A \emph{dyadic cube} is a pair $(I, k)$ of a Borel set $I$ in $X$ and an integer $k \in [-S, S]$ for some fixed natural number $S$. Usually we abuse notation and write $I$ for a dyadic cube $(I, k)$. We write $I \le J$ to mean that for two cubes $(I, k)$ and $(J, l)$, we have $I \subseteq J$ and $k \le l$.
```
grunweg, line 611: I wonder if mentioning the term "order" is still useful. There is later mention of a "top cube", for example.

grunweg (commented):
Thanks! I like these corrections and have just one small comment.

PR #639, opened 2026-08-10
bump to the `v4.33.0` toolchain

lua-vr:
Mathlib will soon bump to the `v4.33.0`, so I also plan to do that here.

I copied the `scripts/runSkimmer.sh` from Mathlib and made it target Carleson; I think it's useful to have it in the future.

The file `Carleson.ToMathlib.Data.Real.ConjExponents` has been upstreamed and is removed.

Sits on top of #637.


fpvandoorn:
Thanks!

PR #645, opened 2026-08-22
feat(push_pr): use lean-action more

austinletson:
With the latest release of `lean-action`, there is a new `mk_all-args` option. This enables using the action for most of CI.


grunweg:
Thanks a lot - both for updating the action and this PR!

PR #649, opened 2026-09-03
paper: edit BoundedCompactSupport section

roos-j:
@fpvandoorn as discussed


fpvandoorn:
This looks good to me. You're stating things a bit stronger than I would, but I don't disagree with anything. I'm also happy to remove the sentences on the degradation on usability.

I'll merge during/after our chat today.

roos-j:
@fpvandoorn I agree actually, I softened it a bit. This can now be merged

PR #650, opened 2026-09-07
chore: more proof reading

roos-j:
- Various minor edits, from a manual and a machine-assisted proof reading. By machine-assisted I mean that I had an AI suggest typo fixes which I then manually reviewed and implemented. 

- Removed a sentence on TauCeti: I propose to not mention any AI-related results that didn't exist at the time the original formalization goal was finished. This opens a large can of worms, which is getting larger by the hour. We can silently rely on the fact that this formalization is "pre-AI".

- Softened some language describing contributors.

PR #651, opened 2026-09-07
Some suggestions after proofreading

jim-portegies:
These are small things that I encountered.

They should probably be considered on a one-by-one basis.


paper/main.tex
```diff
@@ -1482,7 +1483,7 @@ \section{Conclusion}\label{sec:conclusion}
 to formalize foundational results from harmonic analysis,
 which we are submitting to \mathlib
 and which will be useful for the formalization of other results from this field.
-Future projects will also benefit from the insights we learned during the project,
+Future projects will also benefit from the insights we obtained during the project,
```
grunweg: #650 put "gained" here. I don't feel strongly, but think that "gained" is good enough. What do you think?

paper/main.tex
```diff
@@ -377,7 +377,7 @@ \subsubsection{Classical Carleson}\label{subsubsec:classical-result}
     $\lim_{N\to\infty} S_N f(x)$ exists and is equal to $f(x)$.\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/Classical/ClassicalCarleson.lean#L131}{\extlink}
 \end{theorem}
 
-The metric Carleson theorem is more complicated to state, but gives bounds for certain \emph{generalized Carleson operators}.
+The metric space Carleson theorem is more complicated to state, but gives bounds for certain \emph{generalized Carleson operators}.
```
grunweg, line 380: For posterity: prior to this PR, we had four occurrences of "metric Carleson" and nine of "metric space Carleson". I think the latter is slightly better, so :+1: here. Similarly on the same change below.

paper/main.tex
```diff
@@ -839,7 +839,7 @@ \subsection{Stating the main results in Lean}\label{subsec:lean-statements}
...
-\lean{linearizedCarlesonOperator}\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/Defs.lean#L222}{\extlink}; these definitions correspond to those introduced in equations \eqref{def-main-op} and \eqref{def-T-r}, respectively. The expression $\mlean{\|carlesonOperatorIntegrand K (Q x) R_1 R_2 f x\|_e}$ denotes the extended norm of the integral appearing in Equation \eqref{def-T-r}, taking values in $\ennreal$ (see \ref{subsec:enorm} for more details).
+\lean{linearizedCarlesonOperator}\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/Defs.lean#L222}{\extlink}; these definitions correspond to those introduced in equations \eqref{def-main-op} and \eqref{def-T-r}, respectively. The expression $\mlean{\|carlesonOperatorIntegrand \ K \ (Q \ x) \ R_1 \ R_2 \ f \ x\|_e}$ denotes the extended norm of the integral appearing in Equation \eqref{def-T-r}, taking values in $\ennreal$ (see \ref{subsec:enorm} for more details).
```
grunweg: note to self: check if #652 will have an even nicer solution.

paper/main.tex
```diff
@@ -1315,12 +1315,13 @@ \subsection{Working with real numbers}\label{subsec:real}
...
+The choice of real number type also came up when integrating over a subset of real numbers. Before the Carleson project, there was no definition of integration over \lean{NNReal}\href{https://github.com/leanprover-community/mathlib4/blob/81a5d257c8e410db227a6665ed08f64fea08e997/Mathlib/Data/NNReal/Defs.lean#L58}{\extlink} or \lean{ENNReal}\href{https://github.com/leanprover-community/mathlib4/blob/81a5d257c8e410db227a6665ed08f64fea08e997/Mathlib/Data/ENNReal/Basic.lean#L101}{\extlink}. As integrating such functions was useful, we added \lean{volume}\href{https://github.com/leanprover-community/mathlib4/blob/81a5d257c8e410db227a6665ed08f64fea08e997/Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean#L355}{\extlink} instances on these types and proved basic properties and relations. For example, when integrating a function $\mlean{f : \ennreal\to \ennreal}$, the following expressions are mathematically equivalent.
 \begin{lstlisting}[mathescape]
 ∫⁻ (x : ℝ) in Ioi 0, f (ENNReal.ofReal x)
 ∫⁻ (x : ℝ≥0), f (ENNReal.ofNNReal x)
 ∫⁻ (x : ℝ≥0∞), f x
 \end{lstlisting}
+This avoids another source of casting between real number types.
```
grunweg, line 1324: Much better, thanks!

paper/main.tex
grunweg: Yes, it will. Let's merge the other version instead --- but thanks for pointing this one out!

grunweg:
```suggestion
\lean{linearizedCarlesonOperator}\href{https://github.com/fpvandoorn/carleson/blob/020d6e5c71d731952eb0379d222eea4049103ca0/Carleson/Defs.lean#L222}{\extlink}; these definitions correspond to those introduced in equations \eqref{def-main-op} and \eqref{def-T-r}, respectively. The expression $\mlean{\|carlesonOperatorIntegrand K (Q x) R_1 R_2 f x\|_e}$ denotes the extended norm of the integral appearing in Equation \eqref{def-T-r}, taking values in $\ennreal$ (see \ref{subsec:enorm} for more details).
```

grunweg (commented):
Thanks for the comments. I would revert one change (because #652 has a nicer solution), otherwise I agree with all of these.

PR #664, opened 2026-09-23
review remainder of paper

fpvandoorn:
There are some controversial changes here, please review carefully.

* Focused on sections 4.3 and later
* I rewrote a couple paragraphs quite significantly
* Minor changes elsewhere (right-aligned equation numbers, def environment, ...)
* Removed the mention of `filter_upwards`: that seems besides the point for the L^p-design decision. For example, if you want to prove `∫ x, ∫ y, (f + g) (x * y) ∂ν ∂μ = ∫ x, ∫ y, f (x * y) + g (x * y) ∂ν ∂μ`, then most work you have to do is for things unrelated to filter-upwards.
* I removed the section "trade-off between bundled and unbundled objects". I don't see how we can claim that, given that almost all morphisms between algebraic structures are bundled. Feel free to bring it back in some form.
* The `\lean` command was cursed (`\lstinline` reads its argument differently than LaTeX-commands)


paper/main.tex
```diff
@@ -1152,27 +1148,22 @@ \subsection{Standing assumptions and the \texttt{ProofData} pattern}\label{subse
...
-The standing assumptions for the statement of the generalization of
-Carleson's theorem are quite formidable. We will not repeat them here,
-see~\cite[Theorem~1.1.1]{CarlesonBlueprint} for details. Let us just mention
-that they involve a metric space $X$, a measure $\mu$ on $X$ which is
-doubling for a constant $2^a$, a family $\Theta$ of functions from $X$ to
-$\real$, a distance on $\Theta$ satisfying five properties as well as a
-cancellativity condition, and a kernel $K\colon X \times X \to \real$ with
-Calderón--Zygmund like properties. The proofs involve even more additional
-data, notably two bounded measurable sets $F$ and $G$, an integrability
+The standing assumptions for the statement of the generalized
```
mariainesdff, line 1151: I'm not sure about removing these (but I don't have a strong opinion).

paper/main.tex
```diff
@@ -1212,28 +1202,30 @@ \subsection{Standing assumptions and the \texttt{ProofData} pattern}\label{subse
...
+and $\ennreal=[0,\infty]$,\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/Data/ENNReal/Basic.lean#L101}{\extlink}
+which created difficulties in the formalization
 and tension in design decisions.
 Indeed, which of the three types should be used for constants,
 suprema, integrands, and computations?
 
-A term $x$ of type $\nnreal$ is not also of type $\real$, nor is it of type $\ennreal$, even though in set-based mathematics there are inclusions of the corresponding sets. In type theory however, one needs coercions, or casts, to go from one type to another. These coercions are often inserted and applied automatically.
-The \texttt{norm\_cast} tactic can be useful to normalize such casts.
+Since every term has a unique type, we use coercions, or casts, that are automatically inserted to go from one type to another. In particular, there is a coercion from $\nnreal$ to both $\real$ and $\ennreal$.
```
mariainesdff:
```suggestion
Since every term has a unique type, coercions  are automatically inserted to go from one type to another. In particular, there are coercions from $\nnreal$ to both $\real$ and $\ennreal$.
```

paper/main.tex
```diff
@@ -1243,50 +1235,50 @@ \subsection{Working with real numbers}\label{subsec:real}
...
+Yet there is currently no definition for $a^p$ if $p$ is in $\ennreal$,
+and therefore we used conversions from $\ennreal$ to $\real$.
 
 There are also advantages of differentiating between these number types.
 An advantage of putting for instance a constant in \texttt{NNReal} over \texttt{ENNReal} or
 \texttt{Real}, is that it encodes that the constant cannot be infinity and that it needs to be non-negative.
 
 Another advantage of differentiating comes from the different levels of tactic support.
-For real numbers, computations are much easier to perform, using the \texttt{linarith} and \texttt{field\_simp} tactics.
+For real numbers computations are much easier to perform using the \texttt{linarith} and \texttt{field\_simp} tactics.
```
mariainesdff:
```suggestion
For real numbers, computations are much easier to perform using the \texttt{linarith} and \texttt{field\_simp} tactics.
```

paper/main.tex
```diff
@@ -1243,50 +1235,50 @@ \subsection{Working with real numbers}\label{subsec:real}
...
-For both $\ennreal$ and $\nnreal$, it will generally fail when a goal needs properties of subtraction: it even fails to prove that $x - x = 0$. If $x\colon \nnreal$, the \texttt{ring} tactic can prove that $(x^{-1})^2 = (x ^ 2)^{-1}$, or that $3 / x = 3 x^{-1}$, but it fails if $x\colon \ennreal$.
-Similarly \texttt{norm\_num} also works for numbers in $\ennreal$, but again not as well as for real numbers,
-as even with the necessary side conditions it would not simplify a term of the form $a / a$. Both \texttt{ring} and \texttt{norm\_num} fail to solve $(3/5) * (5/3) = 1$ when these numbers are viewed as numbers in $\ennreal$.
+The tactic \texttt{ring} \emph{does} work for these types, but only for operations in semirings (addition, multiplication and numerals).
+In both $\ennreal$ and $\nnreal$, it will generally fail when a goal needs properties of subtraction: e.g.\ it fails to prove that $x - x = 0$.
+In $\ennreal$ it also fails to prove equalities related to division or multiplicative inverses. If $x\colon \nnreal$, the \texttt{ring} tactic can prove that $(x^{-1})^2 = (x ^ 2)^{-1}$ and $3 / x = 3 x^{-1}$, but this fails if $x\colon \ennreal$.
```
mariainesdff:
```suggestion
In $\ennreal$, it also fails to prove equalities related to division or multiplicative inverses. If $x\colon \nnreal$, the \texttt{ring} tactic can prove that $(x^{-1})^2 = (x ^ 2)^{-1}$ and $3 / x = 3 x^{-1}$, but this fails if $x\colon \ennreal$.
```

paper/main.tex
```diff
@@ -1243,50 +1235,50 @@ \subsection{Working with real numbers}\label{subsec:real}
...
+Similarly \texttt{norm\_num} also works for numbers in $\ennreal$, but again not as well as for real numbers. Both \texttt{ring} and \texttt{norm\_num} fail to solve $(3/5) * (5/3) = 1$ when these numbers are viewed as numbers in $\ennreal$.
+Finally, \texttt{field\_simp} fails to simplify a term of the form $x / x$ for $x\colon\ennreal$, even if the necessary side conditions are provided.
 
-On the nose, many results about real numbers transfer to $\ennreal$ provided all arguments are finite; in many cases, these finiteness hypotheses are actually necessary. Manipulating expressions in $\ennreal$ thus often requires proving such finiteness side goals. The \lean{finiteness}\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/Tactic/Finiteness.lean#L73}{\extlink} tactic (initially developed for the formalization of the PFR project\footnote{\url{https://teorth.github.io/pfr/}}) solves many such proof obligations automatically. We made this tactic more powerful, completing (for instance) support for all relevant cases in the Carleson project.
```
mariainesdff, line 1262: I think this flows better keeping the beginning as "On the nose, many results about real numbers transfer to $\ennreal$ provided all arguments are finite; in many cases, these finiteness hypotheses are actually necessary." Otherwise this paragraph has too many stops.

paper/main.tex
```diff
@@ -1243,50 +1235,50 @@ \subsection{Working with real numbers}\label{subsec:real}
...
+A desirable future step is adding improved interaction with the \lean{positivity}\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/Tactic/Positivity/Core.lean#L593}{\extlink} tactic (which solves goals of the form \texttt{$0 \leq x$} or \texttt{$0 < x$}). Proving a goal \texttt{$(x + y + 1)^{-1}<\infty$} (for $x,y\in\ennreal$) requires proving that $x + y + 1$ is non-zero; this is already handled by calling \lean{positivity}. Conversely, \lean{positivity} alone cannot prove goals \texttt{$x^{-1}\neq 0$} for $x\colon\ennreal$.
+Since this is only true when $x$ is finite, this would require that \lean{positivity} calls \lean{finiteness}. This is possible, but is not yet used for division or multiplicative inverses of numbers in $\ennreal$.
```
mariainesdff:
```suggestion
A desirable future step is adding improved interaction with the \lean{positivity}\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/Tactic/Positivity/Core.lean#L593}{\extlink} tactic (which solves goals of the form \texttt{$0 \leq x$} or \texttt{$0 < x$}). Proving a goal \texttt{$(x + y + 1)^{-1}<\infty$} (for $x,y\in\ennreal$) requires proving that $x + y + 1$ is non-zero; this is already handled by calling \lean{positivity}. Conversely, \lean{positivity} alone cannot prove goals \texttt{$x^{-1}\neq 0$} for $x\colon\ennreal$: since this is only true when $x$ is finite, this would require that \lean{positivity} calls \lean{finiteness}. This is possible, but is not yet used for division or multiplicative inverses of numbers in $\ennreal$.
```

paper/main.tex
```diff
@@ -1243,50 +1235,50 @@ \subsection{Working with real numbers}\label{subsec:real}
...
-\footnote{For multiplication, as $0 \cdot \infty = 0$, simplifying $a\cdot\infty$ to $\infty$ involves splitting cases by whether $a$ is zero or non-zero. Lest the reader worry about combinatorial explosion, let us remark that (1) splitting on $a$ being zero presumably implies the case of $a$ being infinite need not be considered separately, and (2) such splits can be performed only as required.}
-The precise design of such a tactic has been discussed (including by some of this paper's authors); its implementation is being worked on.
-% TODO: point to a prototype implementation
+An independently useful and more powerful tactic would allow reducing goals in $\ennreal$ to $\real$: this could be implemented by case-splitting on each relevant variable being finite or infinite. The finite case allows reduction to $\real$ arithmetic; the infinite case will usually be a very simple calculation (as, for example, $\infty$ added to any number in $\ennreal$ is still $\infty$). Such goals can hopefully be proven automatically, or reduced to much simpler tasks for the user.
```
mariainesdff, line 1259: Line 1260 gives the impression that this tactic does not yet exists, immediately contradicted by the following line. Can we clarify this, while still highlighting that this tactic was not available during the Carleson project?

paper/main.tex
```diff
@@ -1243,50 +1235,50 @@ \subsection{Working with real numbers}\label{subsec:real}
...
 \end{lstlisting}
 This avoids another source of casting between real number types.
 
-As a final example around considerations on which number types to use, consider the supremum of a set.
-As one eventually is anyway interested in situations in which this supremum is finite and the set
-is bounded from above, one could consider a setup in which the sets considered are subsets of
-real numbers and the suprema are real numbers as well.
+As a final example, consider the supremum of a set.
+In the Carleson project, we mostly used suprema where we knew the supremum was finite,
+so one could consider a setup where we take suprema over subsets of real numbers
```
mariainesdff:
```suggestion
so we could consider a setup where we take suprema over subsets of real numbers
```
(or use `one` everywhere).

paper/main.tex
```diff
@@ -1243,50 +1235,50 @@ \subsection{Working with real numbers}\label{subsec:real}
...
+and the suprema are real numbers as well.
 This was the choice at the outset of the project.
-However, in this setup, one needs to constantly carry around a proof or assumption that the
-set is indeed bounded.
-This is cumbersome, and therefore it turned out to be much better to
-let suprema just take values in \texttt{ENNReal}.
+However, in this setup, one needs to constantly prove that the set is bounded
```
mariainesdff: Same here.

paper/main.tex
```diff
@@ -1299,27 +1291,28 @@ \subsection{Use of \texttt{ENorm}}\label{subsec:enorm}
...
 
-To illustrate the magnitude of this refactoring, let us give some statistics: phase two (changing the basic definitions) changed about 300 lines of code; phase three (changing lemma statements to \lean{enorm}s) about 1000, and the lemma generalizations so far touched about 2500 lines of code.
+To illustrate the magnitude of this refactoring, let us give some statistics: step two changed about 300 lines of code; step three changed about 3500 so far, with 1000 of these lines consisting of just changing type-classes in lemma statements where the rest of the statement or the proof did not require any further modification.
```
mariainesdff:
```suggestion
To illustrate the magnitude of this refactoring, let us give some statistics: step two changed about 300 lines of code; step three changed about 3500 so far, with 1000 of these lines consisting of just changing type classes in lemma statements where the rest of the statement or the proof did not require any further modification.
```

paper/main.tex
```diff
@@ -1299,27 +1291,28 @@ \subsection{Use of \texttt{ENorm}}\label{subsec:enorm}
...
+The last step is not exhaustive: while substantial parts of \mathlib have been generalized, some areas would require further refactor (and will follow as needed). An example of a refactor\href{https://github.com/leanprover-community/mathlib4/pull/42662}{\extlink} that happened after the Carleson project was completed, was to generalize the total variation of a function enough to apply to functions with $\ennreal$ as codomain. This required generalizing the concept of metric space, since in \mathlib the type $\ennreal$ is not an (extended) metric space. Extended metric spaces require that the topology is generated from the distance, which is not the case in $\ennreal$.
```
mariainesdff:
```suggestion
The last step is not exhaustive: while substantial parts of \mathlib have been generalized, some areas would require further refactor (and will follow as needed). An example of a refactor\href{https://github.com/leanprover-community/mathlib4/pull/42662}{\extlink} that happened after the Carleson project was completed was to generalize the total variation of a function enough to apply to functions with $\ennreal$ as codomain. This required generalizing the concept of metric space, since in \mathlib the type $\ennreal$ is not an (extended) metric space. Extended metric spaces require that the topology is generated from the distance, which is not the case in $\ennreal$.
```
Alternatively, put  "that happened after the Carleson project was completed" between commas (but that seems worse).

paper/main.tex
```diff
@@ -1299,27 +1291,28 @@ \subsection{Use of \texttt{ENorm}}\label{subsec:enorm}
...
+Working with $L^p$ functions presents a further design choice. One option is to work with elements of $L^p$ space, which are equivalence classes of functions up to almost everywhere equality, \lean{MeasureTheory.Lp}\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/MeasureTheory/Function/LpSpace/Basic.lean#L88-L96}{\extlink}. A second option is to work with explicit functions and the condition that they are (a.e.\ strongly) measurable and that the $p$-th power of their norm is integrable, which is defined by the predicate \lean{MemLp}.\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/MeasureTheory/Function/LpSeminorm/Defs.lean#L116-L119}{\extlink}
```
mariainesdff:
```suggestion
Working with $L^p$ functions presents a further design choice: one option is to work with elements of $L^p$ space, which are equivalence classes of functions up to almost everywhere equality, \lean{MeasureTheory.Lp}\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/MeasureTheory/Function/LpSpace/Basic.lean#L88-L96}{\extlink}; a second option is to work with explicit functions and the condition that they are (a.e.\ strongly) measurable and that the $p$-th power of their norm is integrable, which is defined by the predicate \lean{MemLp}.\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/MeasureTheory/Function/LpSeminorm/Defs.lean#L116-L119}{\extlink}
```

paper/main.tex
```diff
@@ -1299,27 +1291,28 @@ \subsection{Use of \texttt{ENorm}}\label{subsec:enorm}
...
-(For example, $(f+g)x = f x + g x$ holds on the nose for actual functions; for representatives of $\tilde{f},\tilde{g}$ and $\tilde{f}+\tilde{g}\in L^p$, a priori this only holds for almost every $x$.)
-While the \lean{filter_upwards}\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/Order/Filter/Defs.lean#L444}{\extlink} tactic simplifies reducing results from almost every to every $x$, these details add up: it is easiest to omit the passage entirely by working with functions in $L^p$ (i.e., with \lean{MemLp} directly).
-Incidentally, lemmas in \mathlib are mostly stated for \lean{MemLp} instead of \lean{Lp}, underscoring this choice.
-
-This relates well with the use of \lean{enorm}s (\S\ref{subsec:enorm}): \mmlean{Lp E p $\mu$} as a normed space requires functions with codomain in a Banach space; this excludes e.g.\ $\ennreal$ which does not possess a well-behaved subtraction operation. Phrasing lemmas using \lean{MemLp} first enables the generalization to $\ennreal$ in the first place.
+For example, $(f+g)(x) = f(x) + g(x)$ holds on the nose for actual functions.
+If we have elements $\tilde{f},\tilde{g}\in L^p$,
```
mariainesdff:
```suggestion
However, if we have elements $\tilde{f},\tilde{g}\in L^p$,
```

paper/main.tex
```diff
@@ -1299,27 +1291,28 @@ \subsection{Use of \texttt{ENorm}}\label{subsec:enorm}
...
+then we can still write down $\tilde{f}(x)$ by picking an arbitrary representative of this class, but then $(\tilde{f}+\tilde{g})(x)=\tilde{f}(x)+\tilde{g}(x)$ only holds for almost every $x$.
+This can make it significantly more cumbersome to manipulate expressions involving elements of the $L^p$ space.
+It is easiest to omit the passage entirely by working with functions with the predicate \lean{MemLp} directly.
```
mariainesdff:
```suggestion
This can make it significantly more cumbersome to manipulate expressions involving elements of the $L^p$ space, so it is easier to directly work with functions with the predicate \lean{MemLp}.
```

paper/main.tex
```diff
@@ -1299,27 +1291,28 @@ \subsection{Use of \texttt{ENorm}}\label{subsec:enorm}
...
+For this reason, lemmas in \mathlib are mostly stated for \lean{MemLp} instead of \lean{Lp}.
 
-This is somewhat similar to the trade-off between bundled and unbundled objects: established common wisdom is that unbundled or partially bundled objects are often better to work with than bundled ones. For instance, a continuous function $f\colon X \to Y$ would usually be formalized as \lstinline|{f : X → Y} (hf : Continuous f)| instead of using the bundled \lean{f : ContinuousMap X Y}.\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/Topology/ContinuousMap/Defs.lean#L33-L37}{\extlink}
+This relates well with the use of \lean{enorm}s (\S\ref{subsec:enorm}): \mmlean{Lp E p $\mu$} as a normed space requires functions with codomain in a Banach space; this excludes e.g.\ $\ennreal$. Phrasing lemmas using \lean{MemLp} first enables the generalization to $\ennreal$ in the first place.
```
mariainesdff, line 1313: "Phrasing lemmas using \lean{MemLp} first enables the generalization to $\ennreal$ in the first place." is confusing.

paper/main.tex
```diff
@@ -1370,10 +1363,9 @@ \subsection{Test function classes}\label{subsec:bdd_comp_supp}
 Staying close to mathematical convention and using everywhere boundedness and ordinary compactness
 pays dividends in formalization, allowing us to provide a richer and more robust API leading
 to simpler proofs.
-Further streamlining is possible by using \lean{fun_prop}.
 
-\lean{fun_prop}\footnote{\url{https://leanprover-community.github.io/mathlib4_docs/tactics.html\#Mathlib.Meta.FunProp.funPropTacStx}}
-is a \mathlib tactic developed by Tomáš Skřivan
+Further streamlining is possible by using \lean{fun_prop}.\footnote{\url{https://leanprover-community.github.io/mathlib4_docs/tactics.html\#Mathlib.Meta.FunProp.funPropTacStx}}
+\lean{fun_prop} is a \mathlib tactic developed by Tomáš Skřivan
```
mariainesdff:
```suggestion
This is a \mathlib tactic developed by Tomáš Skřivan
```

paper/main.tex
```diff
@@ -1428,11 +1420,13 @@ \section{Future work}\label{sec:future}
 Restricted weak-type estimates such as the one in \Cref{metric-space-Carleson}
 specify the boundedness of an operator with respect to the $L^{p,1}$-norm.
 
-There is a generalization of the Marcinkiewicz interpolation theorem to Lorentz spaces,
-which we have not formalized yet.
-However, we have completed a formalization showing that this generalized interpolation result
+There are multiple generalizations of the Marcinkiewicz interpolation theorem that
+could be formalized. Multilinear interpolation, abstract interpolation and interpolation between Lorentz spaces are some examples of this.
```
mariainesdff:
```suggestion
There are multiple generalizations of the Marcinkiewicz interpolation theorem that
could be formalized, including multilinear interpolation, abstract interpolation and interpolation between Lorentz spaces.
```

paper/main.tex
```diff
@@ -1428,11 +1420,13 @@ \section{Future work}\label{sec:future}
...
+An interpolation theorem between Lorentz spaces would be especially useful for this project, since we have completed the formalization showing that this result
 implies \Cref{classical-carleson} for $L^p$ functions with $p>1$.
 
+Another direction is to formalize that \Cref{linearized-metric-Carleson} is strong enough for other applications, such as the almost everywhere convergence of the Walsh--Fourier series for functions in $L^2$.
```
mariainesdff:
```suggestion
Another direction is to formalize further applications of \Cref{linearized-metric-Carleson}, such as the almost everywhere convergence of the Walsh--Fourier series for functions in $L^2$.
```

paper/main.tex
```suggestion
However, in this setup, we need to constantly prove that the set is bounded
```

paper/main.tex
```diff
@@ -1212,28 +1202,30 @@ \subsection{Standing assumptions and the \texttt{ProofData} pattern}\label{subse
 interprets a normed space as a metric space.
 
 The specific implementation of this idea in the Carleson project is slightly
-more involved, as it is written in the form \lean{[ProofData a q K sigma1
-sigma2 F G]}. However, all the additional data $a$, $q$, $K$, $\sigma_1$,
-$\sigma_2$, $F$ and $G$ are inferred automatically from the selected type-class instance (in
-technical terms, they are \emph{outparams}). This means that the user does not
-have to provide them explicitly, as expected of a system designed to
+more involved, as it is written in the form \lean{[ProofData a q K σ₁ σ₂ F G]}. However, all the additional arguments $a$, $q$, $K$, $\sigma_1$,
```
grunweg: These don't look like semantic line breaks to me :-)
Otherwise, this hunk looks good to me.

paper/main.tex
grunweg: I find the paragraph a bit later
> We declared a new type class, of the form [KernelProofData X], containing a measure on X, a doubling
constant, a family Θ of functions from X to R, and so on,

a bit clearer with things repeated. So, also colour me unsure (but not with a hard no).

paper/main.tex
```diff
@@ -1212,28 +1202,30 @@ \subsection{Standing assumptions and the \texttt{ProofData} pattern}\label{subse
...
+and $\ennreal=[0,\infty]$,\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/Data/ENNReal/Basic.lean#L101}{\extlink}
+which created difficulties in the formalization
 and tension in design decisions.
 Indeed, which of the three types should be used for constants,
 suprema, integrands, and computations?
 
-A term $x$ of type $\nnreal$ is not also of type $\real$, nor is it of type $\ennreal$, even though in set-based mathematics there are inclusions of the corresponding sets. In type theory however, one needs coercions, or casts, to go from one type to another. These coercions are often inserted and applied automatically.
-The \texttt{norm\_cast} tactic can be useful to normalize such casts.
+Since every term has a unique type, coercions  are automatically inserted to go from one type to another. In particular, there are coercions from $\nnreal$ to both $\real$ and $\ennreal$.
```
grunweg: The first sentence sounds misleading (if I imagine a non-expert reader). One must still define and register coercions, they don't magically appear. Here's a tentative rephrasing:

Since every term has a unique type, one must distinguish and pass between these real number types explicitly. Often, one can automatically convert from one type to another, and set up the library so these coercions are inserted automatically when necessary.

paper/main.tex
```diff
@@ -1243,50 +1235,49 @@ \subsection{Working with real numbers}\label{subsec:real}
 Consider, for instance, performing a computation on exponents in $L^p$-function spaces.
 Around the real interpolation theorem, one needs to compute $p$ in terms of other exponents that
 naturally live in $\ennreal$.
-Yet there is no definition for $a^p$ if $p$ is in $\ennreal$, and therefore conversions from
-$\ennreal$ to $\real$ are unavoidable.
+Yet there is currently no definition for $a^p$ if $p$ is in $\ennreal$,
```
grunweg: This raises the question "could one change this". Did we discuss this? (Was the answer simply "it's complicated", or just "it would be a lot of work"?)

grunweg: And it seems helpful to mention that a is real. For a an (E)NNReal, this would be easy to define... if it isn't already.

paper/main.tex
```diff
@@ -1243,50 +1235,49 @@ \subsection{Working with real numbers}\label{subsec:real}
...
-as even with the necessary side conditions it would not simplify a term of the form $a / a$. Both \texttt{ring} and \texttt{norm\_num} fail to solve $(3/5) * (5/3) = 1$ when these numbers are viewed as numbers in $\ennreal$.
+The tactic \texttt{ring} \emph{does} work for these types, but only for operations in semirings (addition, multiplication and numerals).
+In both $\ennreal$ and $\nnreal$, it will generally fail when a goal needs properties of subtraction: e.g.\ it fails to prove that $x - x = 0$.
+In $\ennreal$, it also fails to prove equalities related to division or multiplicative inverses. If $x\colon \nnreal$, the \texttt{ring} tactic can prove that $(x^{-1})^2 = (x ^ 2)^{-1}$ and $3 / x = 3 x^{-1}$, but this fails if $x\colon \ennreal$.
+Similarly \texttt{norm\_num} also works for numbers in $\ennreal$, but again not as well as for real numbers. Both \texttt{ring} and \texttt{norm\_num} fail to solve $(3/5) * (5/3) = 1$ when these numbers are viewed as numbers in $\ennreal$.
+Finally, \texttt{field\_simp} fails to simplify a term of the form $x / x$ for $x\colon\ennreal$, even if the necessary side conditions are provided.
```
grunweg: This is implied by the previous sentence that field_simp doesn't support ENNReal. Is the duplication intentional?

paper/main.tex
```diff
@@ -1243,50 +1235,49 @@ \subsection{Working with real numbers}\label{subsec:real}
...
+Yet there is currently no definition for $a^p$ if $p$ is in $\ennreal$,
+and therefore we used conversions from $\ennreal$ to $\real$.
 
 There are also advantages of differentiating between these number types.
 An advantage of putting for instance a constant in \texttt{NNReal} over \texttt{ENNReal} or
 \texttt{Real}, is that it encodes that the constant cannot be infinity and that it needs to be non-negative.
 
 Another advantage of differentiating comes from the different levels of tactic support.
-For real numbers, computations are much easier to perform, using the \texttt{linarith} and \texttt{field\_simp} tactics.
+For real numbers, computations are much easier to perform using the \texttt{linarith} and \texttt{field\_simp} tactics.
```
grunweg: I believe the comma was load-bearing...

paper/main.tex
```diff
@@ -1243,50 +1235,49 @@ \subsection{Working with real numbers}\label{subsec:real}
...
 For terms in $\nnreal$, support in \texttt{linarith} was only added after this project was completed.\footnote{\url{https://github.com/leanprover-community/mathlib4/pull/35155}}
 Neither \texttt{linarith} nor \texttt{field\_simp} support terms in $\ennreal$.
-The tactic \texttt{ring} \emph{does} work for these types, but not as well, and the performance for terms in $\ennreal$ is even worse than for terms in $\nnreal$.
-For both $\ennreal$ and $\nnreal$, it will generally fail when a goal needs properties of subtraction: it even fails to prove that $x - x = 0$. If $x\colon \nnreal$, the \texttt{ring} tactic can prove that $(x^{-1})^2 = (x ^ 2)^{-1}$, or that $3 / x = 3 x^{-1}$, but it fails if $x\colon \ennreal$.
-Similarly \texttt{norm\_num} also works for numbers in $\ennreal$, but again not as well as for real numbers,
-as even with the necessary side conditions it would not simplify a term of the form $a / a$. Both \texttt{ring} and \texttt{norm\_num} fail to solve $(3/5) * (5/3) = 1$ when these numbers are viewed as numbers in $\ennreal$.
+The tactic \texttt{ring} \emph{does} work for these types, but only for operations in semirings (addition, multiplication and numerals).
```
grunweg: "numerals" are not a ring operation, right?

paper/main.tex
```diff
@@ -1243,50 +1235,49 @@ \subsection{Working with real numbers}\label{subsec:real}
...
+The choice of real number type also came up when integrating over a subset of real numbers. Before the Carleson project, there was no definition of integration over \lean{NNReal} or \lean{ENNReal}. As integrating such functions was useful, we added \lean{MeasureSpace}\href{https://github.com/leanprover-community/mathlib4/blob/b63493a4746b4651fceb3bcf3a4651cc36a4b8de/Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean#L355-L356}{\extlink} instances on these types and proved basic properties and relations. Now, the following three expressions to integrate a function $f : \ennreal\to \ennreal$ are equal.
```
grunweg: write "proven to be equal", maybe? They always were mathematically equal - but now we have formal proof, right?

paper/main.tex
```diff
@@ -1299,27 +1290,27 @@ \subsection{Use of \texttt{ENorm}}\label{subsec:enorm}
...
+Pursuing this approach required generalizing many basic measure-theoretic definitions from \mathlib to \lean{enorm}s, as well as all basic results that are required in this project. This was a significant effort, but the verified nature of formalization helped a lot by allowing us to do so incrementally: in a first step, the \lean{ENorm} class was added, together with the enorm instance on normed spaces. Then, all basic definitions were changed to allow an \lean{enorm} as codomain, whenever this did not require further changes. A third step modified many lemma statements to use \lean{enorm}s instead of norms. In this step, some lemmas could be generalized without further modifications, but many lemmas required small changes (such as constants changing from a real to an extended real number, or adding a hypothesis about some quantity in $\ennreal$ being finite). At each point, verification ensured correctness: if all proofs still compiled, no further changes needed to be made.
```
grunweg: I see the intention; this will need some small modifications. Step 3 is really two things
- fixing lemma statements to use enorms,
- generalising e.g. a lemma about lintegral to enorms (which will generalise the codomains, and might not mention enorms explicitly).

Right now, it sounds like it's always the former.

grunweg (commented):
Comments until line 1300

paper/main.tex
grunweg: How about "Phrasing lemmas using \lean{MemLp} (instead of Lp) enables the generalization to functions into \ennreal."?

